PR #231, opened 2011-09-28
view decorator and plugin keywords issue.

fredj:
I'm trying to mix the `@view` decorator with the [bottle-sqlalchemy plugin](https://github.com/iurisilvio/bottle-sqlalchemy/), more or less the same as the attached unit test.

This unit test crash with:
`TypeError: test() takes exactly 1 argument (0 given)`

### DIFF
--- a/test/test_wsgi.py
+++ b/test/test_wsgi.py
@@ -282,6 +282,15 @@ class TestDecorators(ServerTestBase):
         self.assertInBody('401: Unauthorized', '/tpl')
         self.assertStatus(401, '/tpl')
 
+    def test_view_plugin(self):
+        """ WSGI: Test view-decorator with plugin keyword."""
+        @bottle.route('/tpl')
+        @bottle.view('stpl_simple')
+        def test(plugin_keyword):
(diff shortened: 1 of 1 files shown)


cgloeckner:
Is there a chance to continue using the view-decorator while using plugins? I faced the same issue when using the RedisPlugin (bottle-redis).

*app.py*
```python
import bottle
from bottle.ext.redis import RedisPlugin

app = bottle.Bottle()
app.install(RedisPlugin(host="localhost"))

@app.route('/show')
@bottle.view("out")
def show(rdb):
	row = rdb.get("foo")
	if not row:
		bottle.abort(404);
	
	return dict(row=row)

app.run()
```

*out.tpl*
```python
foo = {{row}}
```

Thrown exception: *TypeError: show() missing 1 required positional argument: 'rdb'*

Versions: Redis-Plugin via pip, Bottle via GitHub

benoitm974:
Trying to solve this for the sqlite plugin with 1st level recursion on wrapped function : bottlepy/bottle-sqlite#14

PR #348, opened 2012-06-27
LICENSE.txt renamed to LICENSE in ddb7e5acd

fredj:


prof7bit:
The manifest must be updated also, the pypi package now contains no license file and I have great difficulties using this package in a yocto build, the automated package qa test absolutely requires a license file.

PR #373, opened 2012-09-05
Added abillity to use variables in %include statements

ritave:
Now you can use {{variable}} and {{!variable}} markup in %include statements.

Simple example:
main.py
```python
import bottle

@bottle.route("/")
def index():
    return bottle.template("test", {
            "what":  "template"
        })

bottle.run(host='localhost', port=8080)
```

test.tpl
```
Template 1
%include cool_{{what}}
````

cool_template.tpl
```
Template 2
```


TimLudwinski:
I think with the current version you can use the following syntax to accomplish your goal...

% include("dynamic/" + what, ...)

I suggest closing this ticket.

PR #464, opened 2013-02-25
Minor code fixes and pep8 rules applied

alfonsodg:


oz123:
@defnull  

I think this PR can be closed. It's clear you won't merge it. To many conflicts, since it's too old.

PR #485, opened 2013-04-10
I updated setup.py classifiers to show Python 3.2 and 3.3 support.

ghost:
I added a couple classifiers to show that Bottle works with Python 3.2 and 3.3. 

tox isn't running tests against 3.0 and 3.1 so I didn't mark them as supported.


BubaVV:
Python 3.2 reached EOL status 2 years ago: https://devguide.python.org/#branchstatus . I don't have access to statistics about usage percentage, but 3.2 seems to be very obsolete. Also 3.2 poorly supports venv and testing for it is quite problematic. What do you think about at least dropping 3.2 support from the tox tests?

PR #539, opened 2013-07-30
lookup callable: allows e.g. <template>/de.tpl or modify BaseTemplate.de...

rpuntaie:
...faults per included template or stop an infinite include recursion

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3163,7 +3163,7 @@ class BaseTemplate(object):
                 self.filename = self.search(self.name, self.lookup)
             try: #whether lookup returned generator
                 self.cleanup, self.filename = self.filename, next(self.filename)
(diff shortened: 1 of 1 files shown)


rpuntaie:
I did a rebase. Travis CI fails on some python versions, but so it does for code not related to this pull request. Tell me if I need to do something about it.

eric-wieser:
Don't agree with that. Some failures are definitely due to this pr.

defnull:
2.5, 3.1 and 3.2 fail. A wild guess: Unicode literals? They were introduced in 2.6 and re-introduced in 3.3

rpuntaie:
Corrected: unicode literal u"" was the problem in 2.5, 3.1, 3.2 and AttributeError in 2.5 instead of TypeError in other version.

PR #566, opened 2013-12-19
Fixed path being unshifted before returned generator exhausted.

ergoithz:
I've experienced some request issues on long generator routes, mainly related to SCRIPT_NAME and PATH_INFO changing between yields.
After some debugging I've realized path_shift was being called before yielding, breaking all get_url calls on my partial templates from my mounted app.
This patch fixes that chaining the path_shift to response body, instead of calling right after return.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -650,7 +650,7 @@ class Bottle(object):
         path_depth = len(segments)
 
         def unshift():
-            yield
+            yield ""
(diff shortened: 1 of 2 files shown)


oz123:
@defnull can you close this PR? The OP has deleted it's repo, and there is a cleaner version in #695 .

Thanks, Oz

PR #640, opened 2014-07-23
Fix 602 unicode path encoding when mounting application.

### LINKED ISSUE #602: UTF8 path string invalid when using app.mount()
test.py:

``` python
#!/usr/bin/python
import bottle
import testapp

bottle.debug(True)
app = bottle.Bottle()

app.mount('/test',testapp.app)

app.run(reloader=True, host='0.0.0.0', port=8080)

run(host="localhost",port=8080)
```

testapp.py:

``` python
import bottle

app = bottle.Bottle()

@app.route("/:category", method=["GET","POST"])
def admin(category):
    try:
        return category
    except Exception(e):
        print ("e:"+str(e))
```

Trying to access: http://127.0.0.1:8080/test/äöü results in following error:

```
Error: 400 Bad Request
Invalid path string. Expected UTF-8
```

Running Python 3.4.0 with python-bottle 0.12.5.

blakev:
fix #602

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -831,11 +831,15 @@ class Bottle(object):
 
     def _handle(self, environ):
         path = environ['bottle.raw_path'] = environ['PATH_INFO']
+        
         if py3k:
             try:
                 environ['PATH_INFO'] = path.encode('latin1').decode('utf8')
-            except UnicodeError:
-                return HTTPError(400, 'Invalid path string. Expected UTF-8')
+            except UnicodeError: 
+                try: # BUG 602 encoding path to latin1 may contain char not in utf8
+                    environ['PATH_INFO'] = urlunquote(path) 
+                except: # Really really can't convert the path variable
+                    return HTTPError(400, 'Invalid path string. Expected UTF-8')
 
         try:
             environ['bottle.app'] = self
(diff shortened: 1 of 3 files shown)


onny:
Please merge this into master :(

bottle.py
```diff
@@ -842,11 +842,15 @@ def default_error_handler(self, res):
...
         path = environ['bottle.raw_path'] = environ['PATH_INFO']
+        
         if py3k:
             try:
                 environ['PATH_INFO'] = path.encode('latin1').decode('utf8')
-            except UnicodeError:
-                return HTTPError(400, 'Invalid path string. Expected UTF-8')
+            except UnicodeError: 
+                try: # BUG 602 encoding path to latin1 may contain char not in utf8
+                    environ['PATH_INFO'] = urlunquote(path)
```
mgedmin: This does not seem safe to me.  Repeated UTF-8 decoding might succeed (by accident), mangling actual data.

Besides, the WSGI spec requires environ['PATH_INFO'] to be raw bytes from the HTTP request, decoded as Latin-1.  If you mount a WSGI app, bottle should convert it back before passing it to the mounted app.

IOW I think it would be better to fix this in `mount()`.

defnull:
I think that @mgedmin is right. The actual bug is in `Bottle._handle()`, where `environ['PATH_INFO']` is replaced by the re-encoded value, which breaks the WSGI spec for mounted applications. See #602 for a more detailed explanation.

digitalTranshumant:
So, no solution for this bug?

PR #647, opened 2014-08-06
Add SSL support for built-in WSGIRefServer

njogun:


mfm24:
> - Server adapters are quick-and-simple shortcuts to get a server up and running without any configuration. As soon as you need to configure the back-end, you should no longer use the adapters. They are not flexible enough and not meant to be used in that way.

I didn't realise this. But current SSL support is inconsistent: It's provided only for CherryPy, and isn't using the latest CherryPy API (see #685). To implement SSL using WSGIRef without a server adapter is basically just copying and pasting the existing server adapter followed by a call to ssl.wrap_socket.

How about a pull request where certain server adapters had a method to just get the server (without blocking in serve_forever())? This way instead of having to do the whole server from scratch, it may be possible to do something like

```
s = WSGIRefServer.get_server()
s.socket = ssl.wrap_socket(s.socket, ...)
s.serve_forever()
```

Would this be worth spending time on?

PR #668, opened 2014-10-31
Fix the Allow: header value in 405 "Method not allowed." response.

rgacogne:


sKwa:
+1 (should me `method` and not `verb`).

defnull:
Thanks! Test code coverage is lacking in this area, obviously.

PLPeeters:
@defnull Any reason why this was merged but didn't make it into releases 0.12.14 and later?

PR #670, opened 2014-11-03
Make `root='/'` work cross-platform

eric-wieser:
If `root` already ends in `os.sep`, don't add another one

This used to fail on windows, but somehow worked on unix


eric-wieser:
Ping! Can this be looked at again?

defnull:
Hard to unit-test on non-windows systems, but it looks okay. Thanks

PR #683, opened 2014-11-28
Fix end-of-code-block template token regex

njogun:
If end-of-code-block template token (`%>`) is followed by a Windows-style newline (`\r\n`), bottle will report `SyntaxError: invalid syntax`. 

Example:

``` python
>>> tpl = '<%\r\nprint("hello")\r\n%>\r\nworld'
>>> bottle.template(tpl)
```

Output:

```
  File "bottle.py", line 3494, in template
    return TEMPLATES[tplid].render(kwargs)
  File "bottle.py", line 3320, in render
    self.execute(stdout, env)
  File "bottle.py", line 3307, in execute
    eval(self.co, env)
  File "bottle.py", line 205, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "bottle.py", line 3271, in co
    return compile(self.code, self.filename or '<string>', 'exec')
  File "<string>", line 3
    %>
    ^
SyntaxError: invalid syntax
```

Output after fix:

```
hello
'world'
```

Took me a while to catch that bug as it didn't manifest if end-of-code-block is not followed by a newline (i.e. file ends immediately after `'%>'`).

### DIFF
--- a/test/test_stpl.py
+++ b/test/test_stpl.py
@@ -261,6 +261,9 @@ class TestSTPLDir(unittest.TestCase):
             18
         '''
         self.assertRenders(source, result)
+        source_wineol = '<% a = 5\r\nb = 6\r\nc = 7\r\n%>\r\n{{a+b+c}}'
+        result_wineol = '18'
+        self.assertRenders(source_wineol, result_wineol)
 
     def test_multiline_ignore_eob_in_string(self):
         source = '''


eric-wieser:
Can we fast-track this patch to a release?

defnull:
Released 0.12.10 5 minutes ago :)

eric-wieser:
Thanks!

Part of the issue is that git seems not to recognize `.tpl` as a text format. Perhaps the docs should encourage users to add a `.gitattributes` files containing

    *.tpl text

Which would at least prevent people mistakenly committing windows line endings

PR #685, opened 2014-11-30
Update CherryPy SSL to use latest API and work on Py3

DanielOaks:
On Py2, the `ssl_certificate` and `ssl_private_key` attributes are deprecated, and on Py3 they don't exist at all!

This PR just updates the code to use BottlePy's recommended `ssl_adapter` method, fixing it for Python3.

That said, I haven't actually been able to get this to work with Py2. Neither this nor the previous version work on Python2 for me, and they both result in a non-ssl server, but I suspect that's a CherryPy issue with my particular environment. Mind trying this out with a Py2 server on your own box before pulling it? I've got some test server code [here](https://gist.github.com/DanielOaks/d99607ffda205d6664e2)

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2698,10 +2698,9 @@ class CherryPyServer(ServerAdapter):
             del self.options['keyfile']
         
         server = wsgiserver.CherryPyWSGIServer(**self.options)
-        if certfile:
-            server.ssl_certificate = certfile
-        if keyfile:
-            server.ssl_private_key = keyfile
+        if certfile and keyfile:
+            ssl_adapter = wsgiserver.get_ssl_adapter_class()
+            server.ssl_adapter = ssl_adapter(certificate=certfile, private_key=keyfile)
         
         try:
             server.start()


iLoveTux:
I've been able to get ssl with CherryPy and bottle on Python 2.7.x. I'm not sure why, but I've needed to use CherryPy's BuiltinSSLAdapter (meaning Python's builtin ssl module as opposed to PyOpenSSL [which I don't mind since the Python 2.7.9 update]). The following code creates an ssl server using bottle and CherryPy:

```python
import bottle

app = bottle.Bottle(__name__)

@app.route("/")
def index():
    return "You are seeing this over https."

if __name__ == "__main__":
    from cherrypy.wsgiserver import CherryPyWSGIServer
    from cherrypy.wsgiserver.ssl_builtin import BuiltinSSLAdapter

    CherryPyWSGIServer.ssl_adapter = BuiltinSSLAdapter(
        'cert.pem',
        'privkey.pem',
        None)

    app.run(server="cherrypy")
```

I'm not sure if this will help with your PR and I don't know if this works in Python 3, but I thought I would provide it anyway.

Please Note that there is a bug in CherryPy (I'm not sure about the affected versions) that reports "http" on the CLI when starting the server, but it does serve over https.

mfm24:
I just had this problem, and had to work out if it was a CherryPy or Bottle problem or both. It seems that the latest CherryPy (3.8.0) will happily ignore the ssl_certificate and ssl_private_key members (I don't know if using CPWSGIServer instead of CherryPyWSGIServer would help?). I fixed it in a similar way to iLoveTux.

As a more general question, is SSL actually supported by Bottle? This has been an open pull request  with no comments from the bottle maintainers for over 7 months. If it's not supported why not remove the cherrypy SSL arguments?

PR #690, opened 2014-12-26
Stpl comprehensions

eric-wieser:
Fixes the parser to allow this code to render:

``` python
% a = [
%    (i + 1)
%    for i in range(5)
%    if i%2 == 0
% ]
```

This was written on top of the cleanup in https://github.com/bottlepy/bottle/pull/689

PR #693, opened 2014-12-27
Stpl less slicing

eric-wieser:
Here's another one dependant on #690

The regexes claimed to match "start of line" when in fact, they did not.By using `.search(s, pos=p)` over `.search(s[p:])`, this now works correctly. This allows the special casing of ternary if to be removed.

The previous regex would allow `% if True: foo() end`, the new one requires `% if True: foo(); end`, which seems to be what the old regex intended anyway.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3416,25 +3416,26 @@ class StplParser(object):
                     self.offset += len(line+sep)
                     continue
                 self.flush_text()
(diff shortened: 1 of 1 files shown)


defnull:
It does not speed up the template rendering, but the compiling. I don't expect much performance gain, but the more robust parsing is worth the pull :)

PR #695, opened 2014-12-30
Tidied version of #566

eric-wieser:
#566


defnull:
The get_url() issue should be fixed by now. Mounted bottle applications are now merged internally. The SCRIPT_PATH is not changed, instead the routes are all prefixed with the mountpoint and therefor reflect the complete URL. I added a test case for this and it worked.

PR #699, opened 2015-01-07
Remove unused parameter from the "depr" function.

vascop:
Tried to dig in the commits for a reason for this parameter but couldn't find any. It was once introduced as "hard" with no explanation and then changed to "strict". It never served a purpose as far as I can see.


defnull:
I rewrote the function some time ago and the patch no longer applies. Chances are high that this PR triggered the rewrite, so thanks anyway ;) This can be closed.

Issue #671, opened 2014-11-05
`after_request` won't reflect the out going response on error


grissiom:
Simple test code:

``` .Python
import bottle as bt

@bt.hook('after_request')
def log_request_a():
    print 'resp to: %s, %s' % (bt.request.remote_addr,
                                       bt.response.status_line)

@bt.get('/')
def _get():
    print 'return 400'
    bt.abort(400, 'test')

bt.run(server='cherrypy', host='localhost',  port=1888)
```

It will always print `resp to: 127.0.0.1, 200 OK` , not 400.

stessaris:
Why the changes to fix this issue had been rolled back? In the current version of the code  `after_request` hooks don't receive the error, so the problem is still there.

Is there another way to get hold of the error?

PR #702, opened 2015-01-11
fixes issue #671

vascop:
I'm quite new to Bottle but I've tried to look into this. At [line 861](https://github.com/bottlepy/bottle/blob/master/bottle.py#L861) we get the output of our selected route call. The returned object of `route.call` doesn't update the global `response` object and so when we go into the `finally` block we're dealing with an out of date response.

Furthermore, in case of a 404, we never get to `route.call` and immediately go into the `except HTTPResponse` which also never updates the global response object and so the finally block will get the wrong result.

This updating is done by the `out.apply(response)` method which is called in the `_cast()` method. Unfortunately this is too late because the `after_request` has already been processed.

I don't really know if this is a good solution but I think I've fixed it by creating an `out` variable which gets the updated response object, use the `apply` method in an outer finally block and only then triggering the hook.

Not sure if this is hackish or ugly, but I'm not used to contributing so any feedback would be great.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -852,27 +852,33 @@ class Bottle(object):
             environ['bottle.app'] = self
             request.bind(environ)
             response.bind()
-            try:
-                self.trigger_hook('before_request')
-                route, args = self.router.match(environ)
-                environ['route.handle'] = route
-                environ['bottle.route'] = route
-                environ['route.url_args'] = args
-                return route.call(**args)
-            finally:
-                self.trigger_hook('after_request')
+            out = None
+            self.trigger_hook('before_request')
+            route, args = self.router.match(environ)
+            environ['route.handle'] = route
+            environ['bottle.route'] = route
+            environ['route.url_args'] = args
+            out = route.call(**args)
+            return out
         except HTTPResponse:
-            return _e()
+            out = _e()
+            return out
         except RouteReset:
             route.reset()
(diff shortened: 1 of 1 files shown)


eric-wieser:
Can you produce a testcase that trigger the bad behaviour?

eric-wieser:
All looks fine, but something's breaking the travis build - looks to me like you're causing an error which causes bottle to cause another error when reporting the first error to WSGI, suggesting a bottle bug elsewhere.

eric-wieser:
Somehow bottle's getting a weird `start_response` callback that isn't expecting an `exc_info` argument - try sticking a `print start_response` on [line 978](https://github.com/bottlepy/bottle/blob/master/bottle.py#L978) and seeing where it originates

vascop:
I was calling the apply() without checking. There were exceptions creeping in when the object was a string and so on. It now passes the tests (I should've checked locally with my first pull) and I think it solves the problem. I can try to create a test case for this bug if you think it is necessary.

eric-wieser:
@vascop: There's still an underlying bug in bottle's error handling here, isn't there? The error causing the previous tests to fail was caused by trying to catch your error.

vascop:
Yes, I believe there is another bug here. Not sure if I can fix it though, I don't think I know enough about how bottle works yet. Will probably try to look into it wednesday.

PR #707, opened 2015-01-15
Add SSL/TLS support to TwistedServer

iLoveTux:
First I would like to apologize for any problems related to this pull request, as it is my first pull request on someone else's repository.

The idea here is to add SSL support to the TwistedServer ServerAdapter.

You can see in the comments everything I will say below, but for the sake of completeness:

I try to stick to your conventions as close as possible (I cannot fix the spaces in CherryPyServer that my text editor [NinjaIDE] added). I used a method similar to the one in CherryPyServer to test for certfile and keyfile, and depending on whether they were passed I use the method described [here](http://twistedmatrix.com/documents/12.3.0/core/howto/ssl.html) to have the reactor listenSSL or to listenTCP as the original implementation did.

Please let me know if this pull request adheres to your standards and if not how I can improve for next time. Thank you for your time and consideration.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2787,17 +2787,14 @@ class TwistedServer(ServerAdapter):
         factory = server.Site(wsgi.WSGIResource(reactor, thread_pool, handler))
 
         certfile = self.options.get('certfile')
-        if certfile:
-            del self.options['certfile']
         keyfile = self.options.get('keyfile')
-        if keyfile:
-            del self.options['keyfile']
 
         if certfile and keyfile:
             reactor.listenSSL(self.port, factory,
                 ssl.DefaultOpenSSLContextFactory(keyfile, certfile))
         else:
             reactor.listenTCP(self.port, factory, interface=self.host)
+
         if not reactor.running:
             reactor.run()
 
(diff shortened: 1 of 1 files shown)


eric-wieser:
Don't put "# CHANGE: ilovetux" in the code, that's what git blame is for

iLoveTux:
Ok, thanks. I will fix it as soon as I get to the office this morning. 

Also I noticed that I somehow removed the test to see if The reactor was running. I will fix both of these mistakes.

iLoveTux:
Alright, I removed the comments (all of them because they weren't really needed I had just added them to show what I was changing and why, but as eric-weiser pointed out that is what git blame is for and the comments on my commits cover the rest of the comments which I added to the code).

Also I added back in the check to see if the reactor was running before starting it. I'm still not sure how that had gotten removed (definitely a mistake on my part).

iLoveTux:
I removed the lines which removed keyfile and certfile from self.options because they are not needed.

samstav:
See https://github.com/bottlepy/bottle/pull/647#issuecomment-60152870

PR #708, opened 2015-01-18
Added certfile option to WSGIRefServer to specify SSL

mfm24:
Changes to allow the use of SSL with WSGIRefServer by specifying a certificate file using the 'certfile' key in options.

Based on the example [here](http://www.piware.de/2011/01/creating-an-https-server-in-python/) and my blog post [here](http://www.socouldanyone.com/2014/01/bottle-with-ssl.html). The file specified should be in PEM format containing both private and public keys (it should be possible to add a separate private key too, but this is how I've been using it so far).

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2626,6 +2626,7 @@ class ServerAdapter(object):
         self.options = options
         self.host = host
         self.port = int(port)
+        self.protocol = 'https' if 'certfile' in self.options else 'http'
 
     def run(self, handler): # pragma: no cover
         pass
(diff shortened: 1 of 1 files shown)


kromkrom:
It's a pity this pull request still isn't merged. SSL with default server allows using Bottle in testing to quickly create fully fledged service emulators.

samstav:
See https://github.com/bottlepy/bottle/pull/647#issuecomment-60152870

mfm24:
Thanks samstav, I didn't see that before, I should have searched more.

defnull:
We need to find a good place to document this decision, so no more work is wasted on this topic. Sorry for the difficulty, @mfm24

ghonchesefidi:
Hi guys
I had ran my server in https version but when I request for http version nothing happen.
In other word now that I have https version fine and running, how can I redirect it to https version?
I had seen [this plugin](https://pypi.org/project/Bottle-SSLify/) but I'm using python3.7 and its not working on this version.
Any help?

PR #709, opened 2015-01-19
Add "startup" hook that is called on run()

andrew-d:
The "startup" hook will be called whenever the application is run,
either using `Bottle.run()` or the package-level `run()` function.

In addition, added documentation for the "config" hook.


defnull:
Could you explain a use case for this hook?

PR #711, opened 2015-01-22
Doc translate

avelino:
Configuring environment to have documentation on multi-linguagens


avelino:
@defnull we do merge this and begin to organize the next steps to generate the doc (site) in multi languages?

PR #713, opened 2015-01-31
Allow for long result codes

rcarmo:
When using Hy, all non-floats are implicitly longs, so I keep having to convert everything to int. This is a little fix for that.


eric-wieser:
This fails in py3k because `long` no longer exists

Issue #663, opened 2014-09-23
Plugins are inconsistently applied to @app.error handlers


eric-wieser:
Let's say I have this plugin:

``` python
@app.install
def myplugin(callback):
    def wrapper(*args, **kwargs):
        request.myvalue = True  # actually, some function of session data

        try:
            return callback(*args, **kwargs)
        finally:
            request.myvalue = False

    return wrapper
```

There are 3 scenarios here:
1. **No matching route**
   
    Plugin never executes
    In error handler, `request.myvalue` throws an attribute error
2. **Route matches, raises HTTPError**
   
   Plugin executes, cleanup hit before error handler invoked
   In error handler, `request.myvalue == False`
3. **Route matches, no error**
   
   In route handler, `request.myvalue == True`

I would like `request.myvalue == True` in all three cases.

Can I schedule cleanup to occur after the error handler has executed?
Can plugins be made to run even if no route matches?

PR #717, opened 2015-02-15
fixes #671 and add pass exception to `after_request` hook callback when it it occured.

### LINKED ISSUE #671: `after_request` won't reflect the out going response on error

laiyonghao:
First, `before_request` and `after_request` is very useful hooks for resource management, but `after_request` must know has any exception is raised while request be processing, for decide how to release resource, commit or rollback database transaction for example.
Then, I read @vascop 's PR https://github.com/bottlepy/bottle/pull/702/files and got some inspire. But his code still have some bugs, such as when RouteReset is raise, `after_request` hook callback will be called twice.
So, I extracted some code from _handle() to _inner_handle(). Collected exception and pass it to `after_request` hook callback, if not error, the argument `exc` will be None.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -840,6 +840,19 @@ class Bottle(object):
     def default_error_handler(self, res):
         return tob(template(ERROR_PAGE_TEMPLATE, e=res))
 
+    def _inner_handle(self, environ):
+        try:
+            route, args = self.router.match(environ)
+            environ['route.handle'] = route
+            environ['bottle.route'] = route
+            environ['route.url_args'] = args
+            return route.call(**args)
+        except HTTPResponse:
+            return _e()
+        except RouteReset:
+            route.reset()
+            return self._inner_handle(environ)
+
     def _handle(self, environ):
         path = environ['bottle.raw_path'] = environ['PATH_INFO']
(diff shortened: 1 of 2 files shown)


eric-wieser:
(for inter-issue linking: #671)

eric-wieser:
> before_request and after_request is very useful hooks for resource management

Arguably, that role can also be achieved with `app.install(plugin)`, although there are problems with that too (#663)

laiyonghao:
Yes, Eric, I agree with you.
As you said, `Plugin` is not so good, let's provide more choices to users. :)

defnull:
Is inner_handle used more than once? If not I's inline it again. Function calls and dot lookups are expensive.

PR #721, opened 2015-02-18
Added another recipe for removing trailing slashes

jenmontes:
This one is using a `before_request` hook to strip the trailing slash from `request.environ['PATH_INFO']`. I prefer it to the WSGI middleware method because it leaves the API of the app intact. For example, in the `StripPathMiddleware` snippet, you can't later add a route to `myapp` because you'll get the following error: `AttributeError: 'StripPathMiddleware' object has no attribute 'route'`.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -139,7 +139,7 @@ For Bottle, ``/example`` and ``/example/`` are two different routes [1]_. To tre
     @route('/test/')
     def test(): return 'Slash? no?'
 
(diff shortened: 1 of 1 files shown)


defnull:
The middleware-problem can be fixed this way:

    app.wsgi = StripPathMiddleware(app.wsgi)

but your example is still helpful. merging, thanks :)

jenmontes:
Ah! I had no idea you could apply middleware to just the WSGI! Good to know.

PR #729, opened 2015-03-09
Fix tutorial app routes syntax to bottle 0.10+

avelino:


avelino:
@defnull fixed conflict #741

defnull:
You were quick! I was just about to fix it myself :+1:

PR #736, opened 2015-03-17
only import subprocess if it's needed

mstetson:
Currently, Bottle imports the subprocess module always, even though it's only used with the auto-reloader behavior. This change only imports the module when it will be used.

I'm using Bottle via an embedded Python interpreter in a not-entirely-Python application. Importing the subprocess module causes the embedded Python to try to handle signals. I want all signal handling done in the embedding application, and auto-reload doesn't work for embedding anyway, so I don't want to load subprocess.


defnull:
You have a valid point. Thanks for the patch

PR #739, opened 2015-03-22
.get_undecorated_callback() bugfix

Bystroushaak:
I've found obscure error triggered in case of decorators with multiple parameters.

Now, the ``.get_undecorated_callback()`` method just picks first attribute, which doesn't have to be function, but also one of the parameters of decorator with multiple parameters.

For example, this decorator will trigger the bug:

```python
def database_decorator(fn=None, filename="default.sqlite", table="default"):
    """
    Open connection to the database and offer it to the wrapped function.

    So far, there is no database, only primitive JSON load/store system.
    """
    def database_decorator_wrapper(fn):
        @wraps(fn)
        def database_wrapper(*args, **kwargs):
            path = filename
            if not os.path.isabs(path):
                path = os.path.join(settings.DATABASE_PATH, path)

            # load database
            db = SqliteDict(
                path,
                # autocommit=True
            )

            # load table
            table_ref = db.get(table, {})

            # put table to function parameters
            kwargs["db"] = table_ref

            out = fn(*args, **kwargs)

            # save database
            db[table] = table_ref
            db.commit()

            return out

        return database_wrapper

    if fn:  # python decorator with optional parameters bukkake
        return database_decorator_wrapper(fn)

    return database_decorator_wrapper
```

My patch just looks for cases when the ``func`` is not ``FunctionType`` instance (which it should be) and then goes thru all the parameters and pick first ``FunctionType`` instance.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -42,6 +42,7 @@ if __name__ == '__main__':
 import base64, cgi, email.utils, functools, hmac, imp, itertools, mimetypes,\
         os, re, subprocess, sys, tempfile, threading, time, warnings
 
+from types import FunctionType
 from datetime import date as datedate, datetime, timedelta
 from tempfile import TemporaryFile
 from traceback import format_exc, print_exc
@@ -543,7 +544,16 @@ class Route(object):
         func = getattr(func, '__func__' if py3k else 'im_func', func)
         closure_attr = '__closure__' if py3k else 'func_closure'
         while hasattr(func, closure_attr) and getattr(func, closure_attr):
-            func = getattr(func, closure_attr)[0].cell_contents
+            attributes = getattr(func, closure_attr)
+            func = attributes[0].cell_contents
+
+            # in case of decorators with multiple arguments
+            if func is not None and not isinstance(func, FunctionType):
+                # pick first FunctionType instance from multiple arguments
+                func = filter(
+                    lambda x: isinstance(x, FunctionType),
+                    map(lambda x: x.cell_contents, attributes)
+                )[0]
         return func
 
     def get_callback_args(self):
--- a/test/test_route.py
+++ b/test/test_route.py
@@ -51,7 +51,6 @@ class TestRoute(unittest.TestCase):
             return pretty_dump_wrapper
 
         # multiple decorators
-        @bottle.get("somepath")
         @d2(filename='foo', table='bar')
         @pretty_dump
         def x(a, b):
--- a/test/test_route.py
+++ b/test/test_route.py
@@ -28,3 +28,39 @@ class TestRoute(unittest.TestCase):
         self.assertEqual(route.get_undecorated_callback(), x)
         self.assertEqual(set(route.get_callback_args()), set(['a', 'b']))
 
+    def test_callback_inspection_multiple_args(self):
+        def x(a, b):
+            pass
+
+        # decorator with multiple arguments
+        def d2(fn=None, arg1="something", arg2="something else"):
+            def d(f):
+                def w(*args, **kwargs):
+                    return f()
+                return w
+
+            # support for calling this decorator without arguments
+            if fn:
+                return d(fn)
+
+            return d
+
+        # both decorator parameters
+        route = bottle.Route(None, None, None, d2(arg1='foo', arg2='bar')(x))
+        self.assertEqual(route.get_undecorated_callback(), x)
+        self.assertEqual(set(route.get_callback_args()), set(['a', 'b']))
+
+        # first decorator parameter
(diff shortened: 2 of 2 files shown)


avelino:
Ideally, you write a test to cover implementation (refactoring) proposal.

Bystroushaak:
Well, it didn't break anything, but I will add the tests.

Bystroushaak:
This bug is hard (at least for me) to trigger during test, because it is triggered at runtime.

This is the error message I get for my code:

```python
  File "/usr/local/lib/python2.7/dist-packages/bottle.py", line 862, in _handle
    return route.call(**args)
  File "/usr/local/lib/python2.7/dist-packages/bottle.py", line 1732, in wrapper
    rv = callback(*a, **ka)
  File "/usr/local/lib/python2.7/dist-packages/bottle_gui/bottle_gui.py", line 372, in root
    filter(lambda x: x.path not in BLACKLIST, list_routes())
  File "/usr/local/lib/python2.7/dist-packages/bottle_gui/bottle_gui.py", line 250, in list_routes
    bottle.default_app().routes
  File "/usr/local/lib/python2.7/dist-packages/bottle_gui/bottle_gui.py", line 243, in <lambda>
    args=r.get_callback_args(),
  File "/usr/local/lib/python2.7/dist-packages/bottle.py", line 551, in get_callback_args
    return getargspec(self.get_undecorated_callback())[0]
  File "/usr/lib/python2.7/inspect.py", line 816, in getargspec
    raise TypeError('{!r} is not a Python function'.format(func))
TypeError: './ticker_db.sqlite' is not a Python function
```

when the function is defined as:

```python
DB_FN = "./key_val_db.sqlite"
DB_TABLE = "ticker"

...

@post(SERVICE_URL)
@database_decorator(filename=DB_FN, table=DB_TABLE)
@json_to_params
def save(db, key, val):
    """
    To `key` save `val`.
    """
    db[key] = val
```

Where ``database_decorator()`` was listed in first comment and ``json_to_params()`` is defined here: https://github.com/Bystroushaak/bottle-rest/blob/master/src/bottle_rest/bottle_rest.py#L109

Then when I try to access ``.get_callback_args()`` at **runtime**, the error is triggered. I patched the code with the patch from pull request and the error is fixed.

I am currently trying to write failing test for unpatched version of bottle, but no luck so far.

Bystroushaak:
Ah, finally.

Bystroushaak:
Here is the test body (from ``test_route.py``):

```python
    def test_callback_inspection_multiple_args(self):
        # decorator with multiple arguments
        def d2(fn=None, filename="default.sqlite", table="default"):
            def d(fn):
                def w(*args, **kwargs):
                    kwargs["a"] = filename
                    kwargs["b"] = table
                    return fn(*args, **kwargs)
                return w

            # support for calling this decorator without arguments
            if fn:
                return d(fn)

            return d

        def pretty_dump(fn):
            def pretty_dump_wrapper(*args, **kwargs):
                return fn(*args, **kwargs)

            return pretty_dump_wrapper

        # multiple decorators
        @bottle.get("somepath")
        @d2(filename='foo', table='bar')
        @pretty_dump
        def x(a, b):
            return

        route = bottle.Route(None, None, None, x)
        self.assertEqual(set(route.get_callback_args()), set(['a', 'b']))
```

This fails at ``master`` branch, but not at patched version.

foxbunny:
You should have used a simplified decorator for the test. This one makes the test way too hard to follow.

Bystroushaak:
Simplified decorator doesn't trigger the bug imho. At least I couldnt trigger the bug when I tried.

foxbunny:
No, I mean how is `filename="default.sqlite", table="default"` relevant? Could've been just `foo, bar`. Etc.

Bystroushaak:
Oh, yes. I just cut and pasted parts of my code, where I yestearday found the bug.

Bystroushaak:
I will minimize it and also send the bugfix for python 3, which is now failing in travis.

foxbunny:
You should strip the decorator down to just the parts that exhibit the desired behavior without actually handling args/kwargs, etc, and then test with only that code, so you can rule out any foul in your handling of the args.

Bystroushaak:
The bug shows only when I use kwargs.

foxbunny:
For example, does the bug manifest itself if you **don't** do this:

```
kwargs["a"] = filename
kwargs["b"] = table
```

And does it manifest itself regardless of what the kwargs are named? And so on. Too much noise that doesn't seem too relevant.

Bystroushaak:
Thanks for your feedback. I will try to change the code.

Bystroushaak:
I think, that it will be better, if I change the code and then send new pull request, this one is too long and messy. Sorry.

PR #741, opened 2015-03-23
Don't berate users about their lack of domain-specific knowledge.

colons:
It's [pretty discouraging](http://adamonishi.com/2015/02/obviously-language-in-education/) to be told you should have known something you don't know.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -110,5 +110,5 @@ the unmodified library. In all other cases please ask first.
 
 .. rubric:: Footnotes
 
-.. [1] Usage of the template or server adapter classes of course requires the corresponding template or server modules.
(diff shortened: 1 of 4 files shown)


vaultah:
Perhaps I'm not the one who should respond to you, but I can't see how the use of "of course" berates newbies.

foxbunny:
@vaultah You're not the only one. Though the edits are all ok (they don't take away from the documentation, makes text flow a bit better, etc), the title of the PR is a bit over the top.

avelino:
@foxbunny +1

defnull:
Looks like I really like the term "of course" ;) Thanks for taking the time and rewording these parts.

PR #744, opened 2015-03-24
Removed unnecessary `and func is not None` check.

Bystroushaak:
I couldn't find the case, where func is ``None``.

See #740 for details.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -548,7 +548,7 @@ class Route(object):
             func = attributes[0].cell_contents
 
             # in case of decorators with multiple arguments
-            if not isinstance(func, FunctionType) and func is not None:
+            if not isinstance(func, FunctionType):
                 # pick first FunctionType instance from multiple arguments
                 func = filter(
                     lambda x: isinstance(x, FunctionType),


eric-wieser:
Can't this be simplified further to:

```python
while hasattr(func, closure_attr) and getattr(func, closure_attr):
    attributes = getattr(func, closure_attr)
    func = next(
        attr.cell_contents
        for attr in attributes
        if isinstance(attr.cell_contents, FunctionType)
    )
```

foxbunny:
@eric-wieser Wouldn't that be overoptimizing just for one edge case?

Bystroushaak:
Hmm.

It could, but IMHO it would have worse complexity, because in majority of cases first argument is instance of FunctionType, so the filtering is skipped.

That said, I am not sure of the complexity of generator you are creating, it could be pretty fast and it looks more elegant.

eric-wieser:
@Bystroushaak: The fact it's a generator expression means that `isinstance(attr.cell_contents, FunctionType)` is only evaluated once, which is the same as the `if` statement solution anyway.

@foxbunny: I'd argue it's an improvement, as we've moved the edge case into our normal case - "find the first closure that's a function"

defnull:
I am not happy with this 'magic' anyway. Perhaps it is a better solution to fail hard instead of trying to guess the correct solution in non-obvious cases.

eric-wieser:
Feels like this should be handled by `functools.wraps`. Shame it's not

PR #753, opened 2015-04-24
all_errors wrapper

k3idii:
Feature to cach all errors in one function.
Allow to avoid error(400) error(401) ... declarations.

Normally achived by : 

    def f(*a):
      # code here 
    bottle.Bottle.default_error_handler = f

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -876,6 +876,16 @@ class Bottle(object):
 
         return wrapper
 
+    def all_errors(self):
+        """ Decorator: Register a new default handler for all HTTP errors"""
+
+        def wrapper(handler):
+            self.default_error_handler = handler
+            return handler
+
+        return wrapper
+        
+
     def default_error_handler(self, res):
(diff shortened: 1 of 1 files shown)


eric-wieser:
This would make more sense without the nested function - `@all_errors` vs `@all_errors()`

k3idii:
Well, that is right. Fixed.
Usage now:

    @all_errors
    def handle_error(code):
      return 'Nope'

avelino:
Writting tests pls!

k3idii:
Well, I apologize for my oversight. Im quite new at this (guthib/pullreq/etc/) and still don't know hot to make thinks 'right' way.
Is this sufficient?

        MAGIC = 0x1234
        o = bottle.Bottle()
        @o.all_errors
        def test(*a):
          return MAGIC
        result = o.default_error_handler(None) # just check if user-specific function was called 
        self.assertEqual( result, MAGIC)

jirutka:
+1

PR #767, opened 2015-06-19
Added `code` argument to static_file

SamuelMarks:
Sometimes I use `static_file` to return something I've created during a `POST`. There the `201` response code makes more sense.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2457,10 +2457,8 @@ def _file_iter_range(fp, offset, bytes, maxread=1024 * 1024):
         yield part
 
 
-def static_file(filename, root,
-                mimetype='auto',
-                download=False,
-                charset='UTF-8'):
+def static_file(filename, root, mimetype='auto', download=False,
+                charset='UTF-8', code=200):
     """ Open a file in a safe way and return :exc:`HTTPResponse` with status
         code 200, 305, 403 or 404. The ``Content-Type``, ``Content-Encoding``,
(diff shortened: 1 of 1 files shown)


camconn:
This seems like an unneeded feature.

Why not just do the following? It uses an undocumented method, yet it still works out well. Because `static_file` just returns an `HTTPResponse` object, you can set the response status status manually.
    
```python
static_resp = static_file('some_file.txt', '/some/path')
static_resp._set_status(200)  # 200 can be any other status code
return static_resp
```

eric-wieser:
@camconn: Or maybe avoid the "private" method, and use `static_resp.status = 200`

avelino:
Default source:

    return HTTPResponse(body, **headers)

With this code you see this above just set a status header, example:

    response.set_header('status', 200)

defnull:
Huh? Status is not a header.

SamuelMarks:
Naturally I can modify the response and use the `.status` method, but what's the disadvantage in having it as an argument to `static_file`?

eric-wieser:
> what's the disadvantage

Unpythonic?

`import this`:
* Special cases aren't special enough to break the rules.
* There should be one-- and preferably only one --obvious way to do it.

defnull:
I am with @camconn and @eric-wieser here. `static_file()` returns a `HTTPResponse` that can be modified and changed just fine::

    def new_static_file(*a, **ka)
        res = static_file(*a, **ka)
        if res.status_code == 200:
            res.status = 201
        return res

Issue #769, opened 2015-06-29
Add json property for 'application/json-rpc' requests


kromkrom:
Don't know if there are any pitfalls, but seems to be trivial:
`if ctype in ['application/json', 'application/json-rpc']:`
instead of
`if ctype == 'application/json':`

defnull:
If I remember #616 and understand [1] correctly, browsers white-list allowed content types for cross origin requests (instead of black-listing). So accepting ``application/json-rpc`` should not have any security implications. Right?

[1: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests

kromkrom:
Yes, it seems it makes no difference

oz123:
@defnull, since you merged #773, I guess this issue can be closed.

PR #773, opened 2015-07-08
json for 'application/json-rpc'

kromkrom:
Added json property for requests with 'application/json-rpc' content-type (#769)


camconn:
Minor nitpick: I'd use a tuple instead of a list for checking the mimetype because tuples are a little faster, and they're immutable.

So how about this instead?

```python
if ctype in ('application/json', 'application/json-rpc'):
```

Other than that, looks okay.

kromkrom:
Reasonable!

defnull:
I checked against CVE-2014-3137 and according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Simple_requests this is save. Nice and simple. As soon as more content types are added, we need a more general solution, though.

defnull:
Merged, thanks!

PR #775, opened 2015-07-19
Test of cookie name and value length

netman92:
Added 2 tests of max cookie name and length.


defnull:
Actually the complete cookie line (including options) must not exceed 4k but this is better than no check at all. Perhaps a `len(name)+len(value)` check would do the trick?

PR #778, opened 2015-07-31
Return 400 with bad json request

### LINKED ISSUE #654: Request with invalid JSON throws a 500
Currently, when a request with invalid JSON is sent to the server it causes a `JSONDecodeError` to be thrown by the underlying parser (simplejson in this case).

Bottle should catch and re-throw a 400 Bad Request `HTTPError` to indicated the the client that something is wrong with the request.

jaitaiwan:
JSON requests should not return a 500 error as the issue is with
the requester's json and not the server itself. Fixes #654

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1156,12 +1156,17 @@ class BaseRequest(object):
             property holds the parsed content of the request body. Only requests
             smaller than :attr:`MEMFILE_MAX` are processed to avoid memory
             exhaustion. """
+        err = HTTPError(400, 'Invalid JSON')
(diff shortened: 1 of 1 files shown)


jaitaiwan:
Appears tests failed. No different error messages from master's tests.

defnull:
`JsonDecodeError` is not present in simplejson. Just catching `ValueError` and `TypeError` should do the trick.

jaitaiwan:
Thanks for the feedback. Fixed.

femdom:
Guys, could you, please, release this change?

PR #781, opened 2015-08-17
Add JSON pretty printing support

schlitzered:
Added pretty printing support for JSON responses. This is mainly useful for development, to make it easier to read JSON responses.

The default is to not use pretty printing. But pretty printing can be enabled via:

```
app = Bottle(autojsonpretty=True)
```

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -627,15 +627,17 @@ class Bottle(object):
                          let debugging middleware handle exceptions.
     """
 
-    def __init__(self, catchall=True, autojson=True):
+    def __init__(self, catchall=True, autojson=True, autojsonpretty=False):
 
         #: A :class:`ConfigDict` for app specific configuration.
         self.config = ConfigDict()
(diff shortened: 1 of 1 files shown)


RonRothman:
I see why you'd want this feature, but I vote not to add it.

Bottle's prime directive is to be lean and mean, and this feature can easily and appropriately be implemented as a route decorator (or plugin) instead; or better yet on the client side (Chrome/FF plugin or `python -mjson.tool` on the command line).

schlitzered:
Hey Ron,

i know that this can easily added as a plugin, and indeed it already is. It is only minor enhancement to the JSONPlugin class already bundled with bottle. 

The only overhead added to each response is one more if/else statement.

I hoped it would make more sense to enhance the plugin that is already there, instead of writing a new one & duplicating code.

defnull:
I'm with Ron here. You can archieve the same with `app.install(JSONPlugin(functools.partial(json.loads, ...)))` A recipe for the documentation would be great, though.

PR #783, opened 2015-09-02
make load_app() work on modules in cwd when bottle is called via a script

samstav:
For example, if you call a python script that launches
bottle, it is the path of the script that gets added
to `sys.path`, *not* the current directory. But it is
with respect to the current directory that I would like
to specify the value for `app`:

https://github.com/bottlepy/bottle/blob/0.12.8/bottle.py#L3044-L3045

This makes it so that the application/target can be found,
if it is in the current directory, even if you are
calling another script somewhere else that ultimately
hands over to bottle, and thus the `load_app()` function

The error avoided is:

```python
ImportError: No module named my_app
```

When I execute a script that passes the value `"my_app"`
to `bottle.run(app=value)`

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3148,7 +3148,16 @@ def load(target, **namespace):
         local variables. Example: ``import_string('re:compile(x)', x='[a-z]')``
     """
     module, target = target.split(":", 1) if ':' in target else (target, None)
-    if module not in sys.modules: __import__(module)
+    if module not in sys.modules:
+        try:
+            __import__(module)
+        except ImportError:
+            # cwd is not always in sys.path, e.g when using scripts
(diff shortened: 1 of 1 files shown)


defnull:
This might be dangerous. `cwd` sometimes contains stuff you don't want to import. You can get the same behaviour with `PYTHONPATH=.` or when using bottle.py as a commandline utility or by adding `.` to `sys.path` in your wrapper script. Doing this transparently in `load()` does not feel right.

defnull:
In other words: I think a library should never change `sys.path`. It's the same as registering logging handlers or call `sys.exit()`. Only `__main__` should do that.

samstav:
That's fair. Thanks for reviewing.

Issue #784, opened 2015-09-07
[Bug] ConfigDict.load_config.  'DEFAULT' section not work properly.


kingFighter:
When load config file 
```
[DEFAULT]
interval = 45

[bottle]
port = 8080

[compression]
status = single
```

We incorrectly set  'compression.interval' 

**REASON**
https://github.com/bottlepy/bottle/blob/534a2e08ac0ef55dd542a3a83b1118188c6a399b/bottle.py#L2188-L2192 
*conf.sections* won't return *DEFAULT* section
*conf.items* will return *DEFAULT* options as well.

PR #785, opened 2015-09-07
Add test case for 'load_config'

kingFighter:
#784

### DIFF
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -69,7 +69,34 @@ class TestConfDict(unittest.TestCase):
         c = ConfigDict()
         c.load_dict({key: {'subkey': 'value'}})
         self.assertEqual('value', c[key + '.subkey'])
-   
+
+    @classmethod
+    def setUpClass(self):
+        # create config file
+        from configparser import ConfigParser
+        config = ConfigParser()
+        # two special section
+        config['DEFAULT'] = {'interval': '45'}
+        config['bottle'] = {'port': '8080'}
+        # ordinary section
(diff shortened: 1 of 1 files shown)


test/test_config.py
```diff
@@ -69,7 +69,34 @@ def test_load_dict(self):
...
+    def setUpClass(self):
+        # create config file
+        from configparser import ConfigParser
+        config = ConfigParser()
+        # two special section
+        config['DEFAULT'] = {'interval': '45'}
+        config['bottle'] = {'port': '8080'}
+        # ordinary section
+        config['compression'] = {'status': 'single'}
+        with open('example.ini', 'w') as config_file:
```
samstav: I would use a `tempfile` for this, https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp and use [`TestCase.addCleanup`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.addCleanup) to ensure it is removed when finished. If you use `addCleanup`, you won't need `tearDownClass`.

```python
import functools
import os
import tempfile

...
        # changed from config_file to self.config_file
        self.config_file = tempfile.mkstemp(suffix='.example.ini')
        self.addCleanup(functools.partial(os.remove, self.config_file.name))

        # down in test_load_config()
        c.load_config(self.config_file.name)
```

kingFighter: Really thanks for your advice.

defnull:
I rewrote this in 12445eb. Thanks :)

PR #787, opened 2015-09-11
Fixes #772

### LINKED ISSUE #772: Ability to load config from module
Currently it is not possible to load config from a Python module (similar to the way Django does it).

I'd like to propose the following;

```py
class ConfigDict(dict):
    def load_module(self, module_path):
        """
        Load configuration from module path

        >>> load_module('settings.prod')
        """
        assert isinstance(module_path, str)
        config = importlib.import_module(module_path)
        obj = {key: getattr(config, key) for key in dir(config) if key.isupper()}
        self.update(obj)
```

Happy to put together a tested PR if this proposal is accepted

foxx:
Fixes #772, Includes tests, passed.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2176,14 +2176,20 @@ class ConfigDict(dict):
         self._meta = {}
         self._on_change = lambda name, value: None
 
-    def load_module(self, path):
+    def load_module(self, path, squash):
         """ Load values from a Python module.
-            Example: load_config('my.app.settings')
+            :param squash: Squash nested dicts into namespaces by using
+                           load_dict(), otherwise use update()
+            Example: load_config('my.app.settings', True)
+            Example: load_config('my.app.settings', False)
         """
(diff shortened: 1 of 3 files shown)


samstav:
+1 for Lol in commit messages. Also, nice workaround on dict comprehensions.

LGTM

avelino:
+1
LGTM

defnull:
How about using the bottle.load() function (or its internals) for importing stuff? Thay way you could also load config from a dict in a module (e.g. `myapp.config:conf`) or even from the return value of a function (e.g. `myapp.config:make_config()`).

Also, docs. But that's for later. I'll just merge and we see where we go from here.

foxx:
Ohh, I didn't know about `bottle.load()`, that would be a lot nicer for sure. I'll reopen #772 for improvements. Thanks for the merge/feedback!

PR #794, opened 2015-10-11
cannot import name 'getargspec' in Python 3.6(-dev)

vaultah:
[`inspect.getargspec`](https://docs.python.org/3/library/inspect.html#inspect.getargspec) has been removed

> Deprecated since version 3.0: Use signature() and Signature Object, which provide a better introspecting API for callables. This function will be removed in Python 3.6.

bottle.py fails with the following error:
```
  File "<file>", line 77, in <module>
    from inspect import getargspec
ImportError: cannot import name 'getargspec'
```

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -74,9 +74,28 @@ from types import FunctionType
 from datetime import date as datedate, datetime, timedelta
 from tempfile import TemporaryFile
 from traceback import format_exc, print_exc
-from inspect import getargspec
 from unicodedata import normalize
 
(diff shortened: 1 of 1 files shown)


defnull:
Nice workaround. Thanks

vaultah:
Sorry, I just noticed a small error
```
return (args, varargs, keywords, tuple(defaults) or defaults)
```

should be
```
return (args, varargs, keywords, tuple(defaults) or None)
```

to be consistent with `inspect.getargspec`.

Issue #791, opened 2015-09-20
test_delete_cookie fails with Python 3.5


felixonmars:
Hi, I am getting the following test error when running against Python 3.5:

```
======================================================================
FAIL: test_delete_cookie (test_environ.TestResponse)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/python-bottle/src/bottle/test/test_environ.py", line 614, in test_delete_cookie
    self.assertTrue('name=;' in cookies[0])
AssertionError: False is not true

----------------------------------------------------------------------
```

With Python 3.4 and all other components at the same version, it passes.

kingFighter:
Confirm with `python3.5` under `win8`.
BTW, @felixonmars Can you tell me how to setup `bottle` test environment. 
I couldn't find it in `google`, `http://bottlepy.org/` and  `google groups`

oz123:
@defnull , I know you are not so keen on adding another CI, but https://ci.appveyor.com offers CI for windows. Such issues can than be caught, without having to work directly with Windows (I don't own a Windows Laptop either ...).

defnull:
Cookie test is fixed now in `release-0.12` branch (will be part of next release, whenever that happens).

PR #795, opened 2015-10-11
Fix cookie tests in test_environ.TestResponse

vaultah:
This is a solution for #791.

A few tests are failing because of [some recent changes](http://bugs.python.org/issue23250) in `http.cookies`:
 
 - `test_set_cookie_secure`
 - `test_set_cookie_httponly`

`test_delete_cookie` fails too, I'm not entirely sure why. The workaround is to check for containment of `'Max-Age=-1'`.

Also, I added Python 3.5 to `.travis.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.2"
   - "3.3"
   - "3.4"
+  - "3.5"
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


defnull:
Not the first time python bugfixes in http.cookie broke something. Thanks for the quick fix.

PR #804, opened 2015-11-05
added 'apps' as alias to app, defer 'default app' creation until requ…

team-python:
…ired and add 'default' property to appstack

see https://groups.google.com/forum/#!topic/bottlepy/iMjGmPfJ0tE

This is very minor, tested, and has been discussed on the google group.  Hope all are happy for this.
I have more significant suggestions that can follow.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2323,7 +2323,7 @@ class AppStack(list):
 
     def __call__(self):
         """ Return the current default application. """
-        return self[-1]
+        return self.default
 
     def push(self, value=None):
         """ Add a new :class:`Bottle` instance to the stack """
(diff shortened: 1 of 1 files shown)


defnull:
The 'app stack' is a very old artifact from the very first days of bottle and is far from perfect. The recommended way today is to manually instantiate `app=Bottle()` and not use the stack or the default-application concept at all. At least for multi-application setups. Documentation is lacking in this regard, though.

With the 1.0 release on the horizon we now have the opportunity to re-think these concepts entirely instead of just fixing them up. I'll still pull because the change is minor and 'nice to have'. I still think documentation is more important at this point. A detailed 'how to grow big' tutorial would be extremely helpful.

team-python:
I will work on a detailed 'how to grow big' tutorial, as I am working with an example.  I am still working on how to make building site menu integrate with the app modules.

Issue #616, opened 2014-04-24
json content-type not restrictive enough


stephenR:
Hi,

There is a problem with the content-type parsing that was introduced in #226 .
Bottle will accept things like "text/plain;application/json" as a json content type which might be used by an attacker to bypass security mechanisms.
For example Chrome will not allow cross-origin xmlhttprequests with the content type set to "application/json" but you can set it to "text/plain;application/json" instead and bottle will accept it.

PR #806, opened 2015-11-14
Handle JSON requests with a vendor media type

sloria:
e.g. application/vnd.api+json


defnull:
Are you aware of #616 and checked against it?

defnull:
While it IS save, I cannot merge this. There are to many content-types that use json as as serialization format to support all of them. It is not that hard to build a custom plugin/decorator for these special use cases. Sorry.

PR #808, opened 2015-12-08
Fix to enable async based wsgi server to support yield and yield from in endpoints

SmithSamuelM:
The code in _cast that checks for the content type of iterable interferes with async based wsgi servers where the async endpoints are generators that yield or yield from while waiting for io.

``` python

# Handle Iterables. We peek into them to detect their inner type.
        try:
            iout = iter(out)
            first = next(iout)
            while not first:  # this interferes with async driven endpoints
                first = next(out)
```
Change to
``` python

# Handle Iterables. We peek into them to detect their inner type.
        try:
            iout = iter(out)
            first = next(iout)
           
```

The problem is that this code greedily consumes the iterable until there is non empty content. But it does not need to. An empty unicode string or an empty bytes will still suffice to determine the type.  There may be a use case for an iterable to return an empty string of a different type than what the non empty will be but that seems counter intuitive. The other cases where the itterable returns and HTTPResponse object I don't know about.

Type check still works in the event of an empty string but allows an endpoint to yield an empty string while its async waiting for async io to complete.

For example

``` python
@app.get('/register/<uid>')
def registerGet(uid):
        """
        Example of async endpoint
        """
       # must set content-type before first yield
        bottle.response.set_header('content-type', 'application/json')

        response = yield from myAsyncCoroutine()  # yields "" while not yet completed
        result = json.dumps(response)
        (yield result)

def myAsyncCoroutine():
     while True:
        # Async io here
        if done:
             break
        yield ''  # this is eventually yielded by wsgi app while waiting
     return response
```

I believe this is still compatible with PEP 3333 (see below)as start response has to keep iterating until it gets a
non empty return. This allows the WSGI server to drive the async endpoint coroutine by repeatedly calling the iterater. This allows very simple execution of async wsgi apps. The bottle.run is no longer necessary.

For example the following service routine is repeatedly called by an async scheduler running the the
wsgi http server:

``` python

def service(self):
        """
        Service application
        """
        if not self.closed and not self.ended:
            if self.iterator is None:  # initiate application
                self.iterator = iter(self.app(self.environ, self.start))

            try:
                msg = next(self.iterator)
            except StopIteration:
                self.write(b"")  # if chunked write will send empty chunk to terminate
                self.ended = True
            else:
                if msg:
                    self.write(msg)
                    if self.length is not None and self.size >= self.length:
                        self.ended = True
```
where self.app in this case is a bottle app

With the changes I am able to get my async http server to support yield and yield from in the bottle endpoints where they yield empty strings while waiting for async io operations to complete before returning the response from the endpoint.

From PEP 3333
"However, the start_response callable must not actually transmit the response headers. Instead, it must store them for the server or gateway to transmit only after the first iteration of the application return value that yields a non-empty bytestring, or upon the application's first invocation of the write() callable. In other words, response headers must not be sent until there is actual body data available, or until the application's returned iterable is exhausted. (The only possible exception to this rule is if the response headers explicitly include a Content-Length of zero.)"

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -987,7 +987,7 @@ class Bottle(object):
             environ['PATH_INFO'] = path.encode('latin1').decode('utf8', 'ignore')
 
         def _inner_handle():
-            # Maybe pass variables as locals for better performance? 
+            # Maybe pass variables as locals for better performance?
             try:
                 route, args = self.router.match(environ)
                 environ['route.handle'] = route
@@ -1066,8 +1066,8 @@ class Bottle(object):
         try:
             iout = iter(out)
             first = next(iout)
-            while not first:
-                first = next(iout)
+            #while not first:  # this interferes with async driven endpoints
+                #first = next(iout)
         except StopIteration:
             return self._cast('')
         except HTTPResponse:


SmithSamuelM:
One more thing. This change breaks use of bottle.abort() after the first empty yield. The HTTPError propagates directly into the wsgi server.  The elegant fix for this would be to wrapper the  .__next__ for the iterator  that bottle creates so that if an HTTPError is raised on next() That bottle handles the exception they way it normally would by calling start_response again with exc_info. Since no headers have been output until the first non empty return from the iterator this would be more elegant than the way bottle currently handles HTTPErrors. Which is all front loaded in an if else and assumes that the full response is iterated greedily.  When I get more time I might take a stab at this.

SmithSamuelM:
bottle does not really support the intent of this.

From PEP 3333
"However, the start_response callable must not actually transmit the response headers. Instead, it must store them for the server or gateway to transmit only after the first iteration of the application return value that yields a non-empty bytestring,"

Since it forces a greedy consumption of the iterable. A lazy approach would allow the server to drive the utterable in an async fashion. To do this bottle's error handling of HTTPError needs to be caught in the iterable's .__next__ method not the way it is currently handled.

defnull:
You are right, mostly, but there is no easy solution. Or, more precisely, the proposed solution does not help much. Let me try to explain:

There are two ways for a server to scale: Threads or event loops.

The latter is not well supported by WSGI and does not work with Bottle at all, because event loops handle multiple requests in the same thread context in a non-sequential fashion. Thread-locals (e.g. request and response) will get mixed up. The solution is to either introduce a different form of context-locals (gevent does this and it works very well with bottle), or to not have thread-local state at all. Unfortunately, bottle uses request/response internally. This can be patched out, and I'm working on that, but not without breaking backwards compatibility. An application would not be able to access the request object after the first yield, empty or not, or get a copy first. This is something we actually want to enforce, but only for future versions of bottle.

A threaded server implementation that wants to offload response iterators into a thread pool of workers suffers from the same problem: Thread-locals get mixed up and applications break.

Currently, bottle requires that a request is handled exclusively by the same thread (or greenlet) from start to end, and applications rely on that. In such an environment, blocking the current thread (or greenlet) is the only way to wait for input or output to happen. Passing the response iterable to the server early does not help. The thread is occupied anyway.



But as I said, we are working on this. One idea is to pass request and response as parameters to handlers that wish to be executed asynchronously. This removes the need for thread-locals and fixes most of the problems mentioned above.  It also enables support for 'yield from' and the new 'async' functionality of Python 3.4 by passing in specialized implementations of request/response. If you want to contribute to this discussion, you are very welcomed.

SmithSamuelM:
I appreciate the thoughtful response.  When I get a minute I will post some code to show you more how it works. It seems like I am able to get it to work since I manage the thread local nature of the request response for bottle. I think you need to see the architecture of my coroutine based wsgi server to understand.

SmithSamuelM:
Each bottle app is connected to its own tcp connection so the request response is in that context. So it satisfies your requirement.  "Currently, bottle requires that a request is handled exclusively by the same thread (or greenlet) from start to end, and applications rely on that. In such an environment, blocking the current thread (or greenlet) is the only way to wait for input or output to happen. Passing the response iterable to the server early does not help. The thread is occupied anyway."   So in this case passing the response iterable to the server does work since each iterable is its own coroutine.  The server is servicing the tcp connection so passing control back and forth allows the connection so keep alive and make progress. This is the different between pure coroutines (what I do) and promises/futures/callbacks. I may be missing something but it works for me.

SmithSamuelM:
All I had to do was trap the exceptions that bubble up from bottle because the bottle iterable is not catching them. If the utterable caught them then bottles error handling would be used instead I have to have two types of errors. Those before the first yield and those after the first yield.  I can do yield froms. If the utterable caught the exceptions it would all work.

SmithSamuelM:
Sorry about utterable / iterable. Autocorrect

SmithSamuelM:
I looked at the code.   The threading.local() creates an instance that stores copies of the request and response. So each invocation of bottle from the wsgi server should be ok as its making a copy of the given environment and every invocation has a unique environment.  I am not sure why it would work for gevent and not for the way I am doing it. Maybe that would clarify things for me.

SmithSamuelM:
I think the key is that I have to create a reference to  the bottle.request  .json .query etc as local  variable  in any endpoint before returning control to the server (which I do)  so that subsequent requests do not clobber the copy.  Also because the wsgi server holds  references to the status and response_headers from start_response  it looks like the response local is not stepped on.   So maybe the solution is to inject into each endpoint call a reference to the bottle.request object instead of getting it from the global bottle.request. This local context will be preserved on subsequent yields.  There may be some corner conditions I am unaware of. But I think its close to being able to support fully async coroutine wsgi servers.

SmithSamuelM:
The injection maybe could be done in the _inner_handle closure that wraps the route.call with the environ  so that the route.call has  "request"  and "response" local in scope that the endpoint uses  instead of bottle.request and bottle.response. Otherwise just following the policy of creating a reference request = bottle.request and response =bottle.response  at the top of an endpoint might be enough.

PR #809, opened 2015-12-08
Implemented namespace support for merge

SamuelMarks:


defnull:
I did something similar just recently in e456b4fe389c6410db5c1a813d2991e9f65688ee. Have a look?

SamuelMarks:
@defnull: Hmm… what's the difference between `merge` and `mount`; in terms of use-case; BTW?

defnull:
With e456b4f: Not much.
In general, merge() is meant to combine two applications (probably merging plugins and config, too, in the future) while mount() keeps the application separate.

SamuelMarks:
Hmm, well if there is no longer any difference between the two, how about combining their logic, so that e.g.: `merge` calls `mount`?

/me thought `merge` did something special, with this new information this PR is no longer relevant…

defnull:
There is a difference (mount keeps applications separate, merge merges two apps) or at least there will be a difference once the transition is complete. To maintain BC we have to change these things in small steps.

PR #830, opened 2016-02-06
Mixin args into kwargs earlier in template()

Inndy:
In short: `@view('some_template', template_lookup='./app/my_vews')` won't work

I want to add a lookup-path for templates in my app, so I hook `view` and add `temlate_lookup` to kwargs but it doesn't work.

``` python
from . import bottle
from .bottle import route, request, response

from . import config

view = functools.partial(bottle.view, template_lookup=[config.PATH + '/pages'])
```

This patch fix the problem by mixin args earlier.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3950,6 +3950,8 @@ def template(*args, **kwargs):
     or directly (as keyword arguments).
     """
     tpl = args[0] if args else None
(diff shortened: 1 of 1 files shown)


defnull:
Good find, nice fix. Thanks.

PR #831, opened 2016-02-12
WaitressServer, pass self.options to serve

schlitzered:
pass all options to the WaitressServer adapter, this will allow to set the number of treads via bottle.run


defnull:
Sorry, the pull request is not clean (contains unrelated commits). Please fix that first.

PR #858, opened 2016-06-17
Added support for uvwsgi server

catroot:
Now supporting a fast lubuv based server uvwsgi.


defnull:
From the readme of the project:

"uvwsgi should not be used in production. It's still work in progress.
The project is not very actively maintained at the moment, if you'd like to help please drop me a line!"

Thanks for your pull request but I think we should wait a bit more here.

Issue #859, opened 2016-06-23
static_file() check modified bug


cole-dda:
if a url mapping tow difference file,for examples:
/a/b/c.html => when user is a ,return c1.html
/a/b/c.html => when user is b ,return c2.html

bu if when use static_file(), only use "HTTP_IF_MODIFIED_SINCE"  ，if c2 modified time < c1 modified time , user always see c1.html , must delete browser cache or append random params to url

defnull:
The static_file() helper is for static assets, not for dynamic resources.

You could remove the ``HTTP_IF_MODIFIED_SINCE`` header from `request.environ` to prevent the modification date checking, or just implement the required logic yourself. The static_file() helper does not do much besides last-modified checking and mimetype guessing.

PR #861, opened 2016-07-05
Added ETag support to static_file()

defnull:


defnull:
This should also fix issues like #859 for well-behaved browsers.

PR #862, opened 2016-07-06
port bottle cli optparse to argparse where possible

jheld:
- use argparse in CLI where possible (2.7+)

Note: I'm not sure this should be merged against master...can someone let me know the right merge target if so? And assuming this PR is useful for the project.


defnull:
We dropped support for 2.5 and 2.6 recently. If you are still interested, you could open a new pull request and completely replace optparse with argparse.

PR #864, opened 2016-07-09
remove deprecated logic from GeventServer

eshizhan:
gevent.wsgi just backwards compatibility alias for gevent.pywsgi, `fast` option was useless.
see:
http://www.gevent.org/gevent.wsgi.html
https://github.com/gevent/gevent/commits/v1.1.1/gevent/wsgi.py


avelino:
+1

PR #865, opened 2016-07-14
Add support for flipflop fastCGI connector

ThePrez:


avelino:
Need to write test, so we can evaluate if this working properly
Test server file: https://github.com/bottlepy/bottle/blob/master/test/test_server.py

defnull:
Sorry, but we try to do accept new adapters only if there is a significant demand for it. There are just to many WSGI servers and forks out there to support them all. Since no one commented on this pull request for a while, I'll have to close it. Please feel free to re-open the pull request if the new server adapter matches one of the following conditions:

  * The adapter replaces a popular but abandoned project with an actively maintained fork and solves problems/bugs still present in the original project. In this case, you should also add a deprecation warning to the old adapter.
  * The adapter introduces a new server implementation. In this case, the project should be mature enough for productive use and be supported by an active community.  Python 3 support is required, Python 2 support very welcomed.

PR #872, opened 2016-08-05
Clean python 2.6

oz123:
We are almost 2 years after end of life ...

PEP-361:
Oct 29 2013: Python 2.6.9 final released (security-only)

Python 2.6 should be removed, really.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - "2.6"
   - "2.7"
   - "3.2"
   - "3.3"
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015, Marcel Hellkamp.
+Copyright (c) 2016, Marcel Hellkamp.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ PATH := build/python/bin:$(PATH)
 VERSION = $(shell python setup.py --version)
 ALLFILES = $(shell echo bottle.py test/*.py test/views/*.tpl)
 
-.PHONY: release coverage install docs test test_all test_26 test_27 test_32 test_33 test_34 test_35 2to3 clean
+.PHONY: release coverage install docs test test_all test_27 test_32 test_33 test_34 test_35 2to3 clean
 
 release: test_all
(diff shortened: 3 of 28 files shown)


defnull:
Yes, we can drop official support for 2.6, but we should still test for it to make sure that ALL versions of 2.7 are supported. Debian 7 (LTS and still alive) ships with 2.7.3. Just testing for the newest 2.7 might not catch breaking changes for earlier releases. Python 2.7 is a special case here: 2.6.x releases did not introduce new features, but 2.7.x releases do.

So, either we find a way to test against 2.7.3 or we keep the 2.6 tests to make sure we don't break anything.

oz123:
@defnull , thank for the quick response. 
Unfortunately travis does not allow such fine grainded control... nor tox. But .. I could try and set up testing on https://circleci.com/docs/language-python/ which seem to offer more python versions.

oz123:
Here is a complete list of [python versions supported by circle ci](https://circleci.com/docs/build-image-precise/#python) after 2.6:

-     2.7
-     2.7.3
-     2.7.4
-     2.7.5
-     2.7.6
-     2.7.7
-     2.7.8
-     2.7.9
-     2.7.10
-     3.1.5
-     3.2
-     3.2.5
-     3.3.0
-     3.3.2
-     3.3.3
-     3.4.0
-     3.4.1
-     3.4.2
-     3.4.3
-     3.5.0
-     pypy-2.2.1
-     pypy-2.3.1
-     pypy-2.4.0
-     pypy-2.5.0

When I have some more time I can setup circle ci for this PR. Would you accept it if all tests pass?

oz123:
@defnull I've added a circle.yml which is not really sophisticated, but it gets the job done. All the Python versions are tested. You can see the results here: 
https://circleci.com/gh/oz123/bottle/11

Of course you'll have to add this repostitory to Circle CI too, but I must say it's straight forward and tests run MUCH faster than in Travist (plus, a nice perk, you get ssh access if you need to debug the build).

oz123:
@defnull anychance you will merge this soon? I would like to submit more PRs, and they are related to this one...

defnull:
Merged in branch drop-python-26 and blocked by https://github.com/travis-ci/travis-ci/issues/6525

I really don't want to introduce yet another CI service. We should try to get tests running in travis-ci.org first. If they don't add python 2.7.3, then I'll just build it from source within the travis setup script.

oz123:
Well, travis-ci/travis-ci#6525 is now fixed. Maybe you can restart the build and check again.
BTW, you can still consider adding CircleCI, as a non mandatory build.

defnull:
merged

PR #878, opened 2016-08-29
Remove workaround missing keyword `as`

oz123:
All supported Python versions include the `as` keyword

https://hg.python.org/cpython/file/3.{2,3,4,5}/Lib/keyword.py.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -473,7 +473,7 @@ class Router(object):
             url = ''.join([f(query.pop(n)) if n else f for (n, f) in builder])
             return url if not query else url + '?' + urlencode(query)
         except KeyError as E:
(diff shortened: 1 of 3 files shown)


bottle.py
```diff
@@ -477,8 +472,8 @@ def build(self, _name, *anons, **query):
                 query['anon%d' % i] = value
             url = ''.join([f(query.pop(n)) if n else f for (n, f) in builder])
             return url if not query else url + '?' + urlencode(query)
-        except KeyError:
-            raise RouteBuildError('Missing URL argument: %r' % _e().args[0])
+        except KeyError as E:
```
eric-wieser: Shouldn't this be lower case?

oz123: You mean `E` ?

PR #880, opened 2016-08-30
Add documentation for ConfigDict.load_module

### LINKED ISSUE #772: Ability to load config from module

oz123:
Fix #772

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2257,25 +2257,29 @@ class ConfigDict(dict):
         self._fallbacks = []
 
     def load_module(self, path, squash=True):
-        """ Load values from a Python module.
+        """Load values from a Python module.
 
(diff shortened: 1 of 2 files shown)


defnull:
Great, thanks!

PR #884, opened 2016-09-02
Re enable coverage

oz123:
Enable coverage reporting and collection in coveralls.


defnull:
coverage.py is broken for Python 3.2. I think it is enough to test coverage with 3.5 and 2.7, but the results should be combined. I'm not sure how to do that with travis.

oz123:
@defnull , I found a fix for it.

PR #889, opened 2016-09-09
Fix warning about using absolute template path names

### LINKED ISSUE #888: Using Jinja2Template produces deprecation warnings
Using code like this
```python
@app.get('/hello')
def hello():
    return template('hello.html', template_adapter=bottle.Jinja2Template)
```
causes a deprecation warning to be emitted:
```
.../bottle.py:3325: DeprecationWarning: Absolute template path names are deprecated.
  fname = self.search(name, self.lookup)
```

mgedmin:
Fixes #888.


defnull:
Thanks!

PR #890, opened 2016-09-09
Drop opt parser

oz123:
Another tiny pull request ...

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -20,16 +20,15 @@ __version__ = '0.13-dev'
 __license__ = 'MIT'
 
 ###############################################################################
-# Command-line interface ########################################################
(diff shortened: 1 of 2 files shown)


oz123:
The latest build actually passed. Here is a correct link:

https://travis-ci.org/oz123/bottle/builds/158829658

See also issue #891 .

You should seriously consider moving the CI to CircleCI. Buillds there are much more flexible, faster and there is SSH. You can access the build environment in case stuff breaks, and see why stuff breaks.

defnull:
The server adapter tests are highly unstable on Travis-CI. We can try Circle-CI and consider a switch. I just logged in there and gave Circle-CI the required permissions.

Thanks for your work!

PR #893, opened 2016-09-21
Fix tracebacks that would contain <string> to contain the template path

eric-wieser:
It seems that using `eval(compile(...))` throws away the name that we gave the `code` object. We don't use the return value of `eval` anyway, so `exec` is fine.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -952,7 +952,7 @@ class Bottle(object):
         return wrapper
 
     def default_error_handler(self, res):
-        return tob(template(ERROR_PAGE_TEMPLATE, e=res))
+        return tob(template(ERROR_PAGE_TEMPLATE, e=res, template_settings=dict(name='__ERROR_PAGE_TEMPLATE')))
 
     def _handle(self, environ):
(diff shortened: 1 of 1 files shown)


eric-wieser:
Oops, `exec` is a function in py3k

eric-wieser:
Not actually sure if it was even broken any more, but either way `exec` is more correct.

bottle.py
```diff
@@ -952,7 +952,7 @@ def wrapper(handler):
         return wrapper
 
     def default_error_handler(self, res):
-        return tob(template(ERROR_PAGE_TEMPLATE, e=res))
+        return tob(template(ERROR_PAGE_TEMPLATE, e=res, template_settings=dict(name='__ERROR_PAGE_TEMPLATE')))
```
eric-wieser: Open to a better name here. Should be something that's clearly the default template handler, but not going to clash with any names in userspace

eric-wieser:
Test failures look like garbage to me

defnull:
Yes, the travis-ci tests are broken and circle-ci are not configured jet. Working on it :) Thanks for the patch!

PR #894, opened 2016-09-21
Handle unprintable errors

eric-wieser:
Previously, these would throw another error and obliterate the useful traceback, replacing it with a useless one

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -4112,7 +4112,12 @@ ERROR_PAGE_TEMPLATE = """
             <pre>{{e.body}}</pre>
             %%if DEBUG and e.exception:
               <h2>Exception:</h2>
-              <pre>{{repr(e.exception)}}</pre>
+              %%try:
+                %%exc = repr(e.exception)
+              %%except:
+                %%exc = '<unprintable %%s object>' %% type(e.exception).__name__
+              %%end
+              <pre>{{exc}}</pre>
             %%end
             %%if DEBUG and e.traceback:
               <h2>Traceback:</h2>


defnull:
Out of curiosity, which exception raises an exception when printed? Isn't that a serious bug in the custom exception class itself?

eric-wieser:
Here's a (contrived) example::

```python
class MyBadObject:
    def __repr__(self):
        raise Exception('Something went wrong getting the repr')
```

In a normal python shell:
```python
>>> data = {}
>>> o1 = MyBadObject()
>>> data[o1]
>>> class MyBadObject:
    def __repr__(self):
        raise Exception('Something went wrong getting the repr')

>>> data = {}
>>> bad = MyBadObject()
>>> data[bad]
  File "<pyshell#7>", line 1, in <module>
    data[bad]
KeyError: <unprintable KeyError object>
```
Whereas in bottle, there is no mention of a `KeyError`

defnull:
Ahh, make sense :)

eric-wieser:
The real issue was that `MyBadObject.__repr__` incurred database accesses on lazy properties, but by the time the error handling template is reached and `repr` is called, the database session has already been closed.

eric-wieser:
Actually, is there an argument here for making `HttpError` cache `repr(exception)` on construction, which would nicely fix the lifetime issue I was seeing?

defnull:
I tend to think that triggering database queries in ``__repr__`` is bad practice and raising exceptions there is just wrong. Whoever calls repr(obj) cannot possibly know about the hidden DB calls or implementation details of the DB layer and would not be able to handle them correctly, anyway. The DB layer should implement ``__repr__`` in a more robust way and handle its own exceptions.

PR #899, opened 2016-10-08
Static file multithreading

niccokunzmann:
Context: When I use `bottle.run(..., server='cherrypy')`, I run it in
multi-threaded mode. This results in two requests sharing the same
environment.
They serve the old file with 200 in this case.
Fix: Getting the header from the request object first, makes the actual
request being used and not the request that runs in parallel.
Evaluation: This results in a performance improvement in this scenario,
since one more file is reported as not modified.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2835,7 +2835,9 @@ def static_file(filename, root,
         if check and check == etag:
             return HTTPResponse(status=304, **headers)
 
-    ims = getenv('HTTP_IF_MODIFIED_SINCE')
+    ims = request.get_header('If-Modified-Since')
+    if not ims:
+        ims = request.environ.get('HTTP_IF_MODIFIED_SINCE')
     if ims:
         ims = parse_date(ims.split(";")[0].strip())
     if ims is not None and ims >= int(stats.st_mtime):


defnull:
> Context: When I use bottle.run(..., server='cherrypy'), I run it in
multi-threaded mode. This results in two requests sharing the same
environment.

This should never happen. ``bottle.request.environ`` is a thread-local attribute and all other attributes of ``bottle.request`` are referenced from this thread-local environment dictionary. If there are mix ups, then something serious is going wrong and fixing ``static_file`` is not enough. A WSGI server should never re-use the same environ dictionary for a different request, nor should it change the thread-context of a request while it is handled. The behavior you are seeing is not normal.

> Fix: Getting the header from the request object first, makes the actual
request being used and not the request that runs in parallel.

Your patch should not be able to solve the issue, because ``request.get_header()`` accesses ``request.environ`` internally. It should not make any difference.

The only explanation I have is that your setup of cherrypy uses greenlets or a different non-standard form of threads and the monkey-patching of the standard-lib was incomplete while importing bottle.

Can you provide a minimal runnable test setup?

defnull:
Closed because I could not reproduce the issue. Please reopen with a minimal test scenario or test case.

Issue #841, opened 2016-03-29
Explicitly flush wsgi.errors after writing errors


amluto:
uwsgi is impolite and supplies a block-buffered object for wsgi.errors.  The result is that exception traces get lost or severely delayed.

Please consider adding a call to environ['wsgi.errors'].flush() after all the writes to wsgi.errors to force the errors out to the log.

PR #903, opened 2016-10-19
Flush wsgi.errors

sirex:
Some wsgi servers buffers `wsgi.errors` and error messages does not appear in logs. To fix that, after writing to `wsgi.errors` it should be flushed in order for content to be written to the file.

See: https://github.com/unbit/uwsgi/issues/1222


sirex:
This should fix #841.

PR #917, opened 2017-01-04
New plugins proposal

kianxineki:
I have created two new plugins:

## Bottle-errorsrest:
returns all errors in json format (even those generated with HTTPError)

## Bottle-cerberus:
Integrates cerberus schemas into bottle decorators


avelino:
thank you @kianxineki

PR #920, opened 2017-01-05
Added to new Plugins

PyB1l:
- bottle-jwt: JSON Web Token authentication plugin for bottle.py
- bottle-smart-filters: Bottle Querystring smart guessing.


PyB1l:
Both plugins are useful for building web services

defnull:
We have two jwt plugins now. One should probably go.

PyB1l:
Is there something wrong with pull request? Should I make another one?

PR #924, opened 2017-01-09
that should fix #922

### LINKED ISSUE #922: The wget installation to wget http://bottlepy.org/bottle.py raise 404 error
When following the tutorial wget wget http://bottlepy.org/bottle.py installation we get a 404.

simevo:


defnull:
I would rather fix the server setup, since these short links are quite handy if I want to download bottle without venv or pip installed.

defnull:
I'll do that as soon as I get home.

PR #936, opened 2017-02-02
It's 2017....

wogsland:


defnull:
@eric-wieser is technically right, but I'll still pull this ;)

PR #940, opened 2017-02-08
Httptools intergration

Benozo:
Trying to make Bottle run faster as per req/sec. So far on my machine running apache bench its pretty fast. 
Faster then any other ServerAdapter.

running 'ab -n 100 -t 10 http://<SERVER-IP>:8081/' and returning simple JSON data.

When benched against Sanic, Bottle was actually on par with Sanic. **Note**: Someone should bench it and see if this statement is true or not, because it could just be my machine doing weird things...


kianxineki:
File "...bottle.py", line 3667
    async def async_process_response(self):
            ^
SyntaxError: invalid syntax

Python 3.4.2

It would be necessary to support earlier versions of python

defnull:
Also, this pull request contains a lot of unnecessary commits and changes. Please open a new pull request once the issues are resolved. Thanks

Benozo:
@kianxineki not possible with Httptools . Httptools only support Python 3.5+ . I guess this is the limitations of Httptools and uvloop .@defnull I will make.a clean PR sometime this week .

defnull:
Python 2.7 compatibility is a requirement for any code that is integrated into the core framework. You can publish your server adapter as a separate plug-in, though.

kianxineki:
I mean the same thing you said @defnull forgive for my English!

PR #943, opened 2017-02-20
Update .travis.yml

edmBernard:
@defnull fix travis problem with coverage and python 3.2
remove `pip install -U coverage`
after `pip install coverage==3.7.1` in case of python 3.2


edmBernard:
It don't seem to fix correctly the issues.
It only silent problem with coverage

defnull:
Your other PR helped a lot to solve this. Thanks.

PR #944, opened 2017-02-20
try fix travis 

edmBernard:
@defnull 
remove `pip install -U coverage` after after python version check


edmBernard:
failed without all your change prefer but the branch fix travis worked

Issue #934, opened 2017-02-02
Update CherryPy support for v 9.0.0 +


RichardLangFromNZ:
Current bottle (v0.12.13) is failing to import cherrypy.wsgiserver on line 2787 as it no longer exists.

[]http://docs.cherrypy.org/en/latest/history.html#v9-0-0

Work around by installing cherrypy version < 9.0.0

mariolong:
I install `libev` on archlinux.
it is work for me.

jasondrew:
A better solution is this fix that I applied inside bottle.py. There's still a wsgiserver implementation available in cherrypy, it's just moved/renamed into 'cheroot'.

`class CherryPyServer(ServerAdapter):
    def run(self, handler):  # pragma: no cover
        # JDrew: added the 2nd import to handle the latest cherrypy not
        # supporting 'from cherrypy import wsgiserver'.
        try:
            from cherrypy import wsgiserver
        except ImportError:
            from cheroot.wsgi import Server as wsgiserver`

liffiton:
This appears to have been fixed in be9081411 and nearby commits, but this particular issue wasn't noted in the commit message.  I expect the next release will include that support for ``cheroot``.

dimyme:
line 3271 fixed for me: 

```
        # from cherrypy   import           wsgiserver                                  # This will fail for CherryPy >= 9                                   ###    TODO
        from cheroot.wsgi import Server as WSGIServer
        self.options['bind_addr'] = (self.host, self.port)
        self.options['wsgi_app'] = handler

        certfile = self.options.get('certfile')
        if certfile:
            del self.options['certfile']
        keyfile = self.options.get('keyfile')
        if keyfile:
            del self.options['keyfile']

        server =              WSGIServer(**self.options)  ################# ###     wsgiserver.CherryPyWSGIServer(**                  WSGIServer.CherryPyWSGIServer(**

```

PR #947, opened 2017-02-23
Leave explicit the maxima version supported the CherryPy (<= 9.0.0)

avelino:
ref #934


edmBernard:
to replace cherrypy by cheroot : https://github.com/bottlepy/bottle/pull/932

Issue #951, opened 2017-03-10
Setting bottle.BaseRequest.MEMFILE_MAX does not affect tornado.httpserver.HTTPServer max_body_size 


gf-atebbe:
I'm trying to send a multipart/form-data request to a bottle service that is using the tornado server.  I set bottle.BaseRequest.MEMFILE_MAX=1048576000 in my instance and bottle does not complain about the size of the request.  However, the size of the body exceeds tornado's default value of 104857600, so tornado returns a HTTP 413 error

```
Malformed HTTP message from 172.17.0.1: Content-Length too long
```

Is there any way to make bottle.BaseRequest.MEMFILE_MAX propagate through to the instance of tornado.httpserver.HTTPServer that is instantiated in bottle.TornadoServer?

```python
class TornadoServer(bottle.TornadoServer):
    """ The super hyped asynchronous server by facebook. Untested. """
    def run(self, handler): # pragma: no cover
       import tornado.wsgi, tornado.httpserver, tornado.ioloop
       container = tornado.wsgi.WSGIContainer(handler)
       server = tornado.httpserver.HTTPServer(container) # <- Line to change below
       server.listen(port=self.port,address=self.host)
       tornado.ioloop.IOLoop.instance().start()
```

Passing the max_body_size when instantiating tornado.httpserver.HTTPServer resolves my issue:
```python
server = tornado.httpserver.HTTPServer(container, max_body_size=1048576000)
```

Is there any way to do this from bottle's interface?

kianxineki:
The same happens with the meinheld server 👍

defnull:
`MEMFILE_MAX` limits the maximum amount of data hold in memory buffers instead of temporary files. It does not limit the total size of the body part. If a request is larger than `MEMFILE_MAX` it is simply buffered into a file instead of an ``io.BytesIO`` to prevent out of memory errors.

So, while it is a good idea to make `max_body_size` configurable, `MEMFILE_MAX` is certainly the wrong value.

PR #952, opened 2017-03-14
MAX_BODY_SIZE support in tornado and meinheld

kianxineki:
Issue #951

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1142,6 +1142,9 @@ class BaseRequest(object):
 
     __slots__ = ('environ', )
 
+    #: Maximum size of body in bytes.
+    MAX_BODY_SIZE = 102400
+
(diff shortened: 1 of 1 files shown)


defnull:
Since `max_body_size` is a configuration of the server backend (and not bottle) I would not put it in `BaseRequest` as a constant. Other than `BaseRequest.MEMFILE_MAX` which is just a more or less arbitrary constant and does not affect application behavior (only performance), `max_body_size` has a very visible effect and should not be hidden in an undocumented class-constant. 

The `run()` method passes any additional parameters to the server backend adapter, and some adapters already support additional parameters to configure the backend appropriately. How about supporting `run(server='meinheld', max_body_size=...)` and potentially other parameters as well?.

PR #958, opened 2017-03-18
make app.error can be used as a decorator or called directly.

iamybj:
""" Register an output handler for a HTTP error code.
            Can be used as a decorator or called directly ::
                
                def error_handler_500(error):
                    return 'error_handler_500'

                app.error(code=500, callback=error_handler_500)

                @app.error(404)
                def error_handler_404(error):
                    return 'error_handler_404'
        """

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -948,18 +948,18 @@ class Bottle(object):
         return self.route(path, method, **options)
 
     def error(self, code=500, callback=None):
-        """ Register an output handler for a HTTP error code.
-            Can be used as a decorator or called directly ::
-                
+        """ Register an output handler for a HTTP error code. Can
+            be used as a decorator or called directly ::
+
                 def error_handler_500(error):
                     return 'error_handler_500'
 
                 app.error(code=500, callback=error_handler_500)
 
-
                 @app.error(404)
                 def error_handler_404(error):
                     return 'error_handler_404'
+
         """
         
         def decorator(callback):
--- a/bottle.py
+++ b/bottle.py
@@ -947,14 +947,27 @@ class Bottle(object):
         """ Equals :meth:`route` with a ``PATCH`` method parameter. """
         return self.route(path, method, **options)
 
-    def error(self, code=500):
-        """ Decorator: Register an output handler for a HTTP error code"""
+    def error(self, code=500, callback=None):
+        """ Register an output handler for a HTTP error code.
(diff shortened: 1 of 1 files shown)


iamybj:
why still not merged?

eric-wieser:
Why is this desirable? We can already do `app.error(code=500)(error_handler_500)`. Do `app.route`, `app.post`, etc support this behaviour?

A good PR explains why and not just what.

defnull:
Yes, `app.route`, `app.post` etc. do accept the callback as a parameter if you do not want to use  decorator syntax. This patch would make `app.error` behave more like `app.route`.

A test case would be nice. Other than that, I don't see any reasons not to pull this request. It's small enough and does no harm.

bottle.py
```diff
@@ -947,14 +947,27 @@ def patch(self, path=None, method='PATCH', **options):
...
+                app.error(code=500, callback=error_handler_500)
+
+                @app.error(404)
+                def error_handler_404(error):
+                    return 'error_handler_404'
+
+        """
+        
+        def decorator(callback):
+            if isinstance(callback, basestring): callback = load(callback)
```
eric-wieser: Ah, consistency with `route`

eric-wieser:
Might be nice to provide a decorator to automate this optional-decorator behaviour. Something like:

```
def callback_decorator(f):
    @functools.wraps(f)
    def wrapped(f, *args, **kwargs):
        callback = kwargs.get('callback', None)
        if callback is None:
            def decorator(callback):
                return f(*args, callback=callback, **kwargs)
            return decorator
        return f(*args, **kwargs)
    return wrapped
```

iamybj:
decorator routing will make the route instructions spread all over the project, it is not good design pattern. I like hard code style routing:
`
app = Bottle()

app.error(code=500, callback=error_handler)

app.get(path='/static/<path:path>', callback=functools.partial(static_handler, root='wwwroot/static/'))
app.get(path='/', callback=home_handler)
`

bottle.py
iamybj: yes, I copy the load method from route

eric-wieser:
> decorator routing

I'm not talking about that. My point was that the _implementation_ of `route` and `error` could share an implementation by using an internal decorator

iamybj:
yes, any solution is ok, I just need this feature.

Issue #982, opened 2017-07-13
Add SameSite support for cookies


hyde0620:
It would be great if we can set "**SameSite**" flag via set_cookie function.

e.g. 
response.set_cookie(cookie_key, cookie_id, same-site-cookies='Lax', ....)

provinzkraut:
For a workaround, you could just add it via `add_header()` like this:
`response.add_header('Set-Cookie', 'cookie_key=cookie_id; SameSite=Lax')`.

But indeed it would be nice to have this as a keyword for `set_cookie()`.

provinzkraut:
However, if you want to do this via `set_cookie` you could modify your bottle.py like this:

import Morsel from http.cookies: `from http.cookies import Morsel`.
Then add the following to `set_cookie()`:
`Morsel._reserved['same-site'] = 'SameSite'`.

Now you can use the attribute `same_site` in `set.cookie()`, like
`response.set_cookie(cookie_key, cookie_id same_site="lax")`.

This updates the `_reserved` dict, which is used to check if an attribute is valid.

hyde0620:
@provinzkraut  Thanks for the reply. 
It's a good way to add the attribute `same_site`. 👍

oz123:
@defnull this issue is resolved.

Teva:
Hi all,

Am I wondering if my issue is linked to this closed issue :

```python
 response.set_cookie(name = "name", value = name, secret='some-secret-key', httponly = True, same_site = 'strict')  #secure = True to be added in prod
```

this raises the following error with python 3.7 :
MYPATH--> /http/cookies.py", line 311, in __setitem__
    raise CookieError("Invalid attribute %r" % (K,))
http.cookies.CookieError: Invalid attribute 'same-site'

susundberg:
I ended up hitting the excatly the same error message with bottle 0.12.18 (Python 3.6.9)

defnull:
The fix (or same-site support) was not back-ported to 0.12 yet.

PR #983, opened 2017-07-13
"SameSite" cookie feature (from issue #982)

provinzkraut:
As mentioned in #982, `set_cookie()` does not support  the "SameSite" cookie attribute.
This is because the `http.cookies` module doesn't support that either.
A reason for _this_ isn't really given. It's just that the module was based on RFC 2109 which "SameSite" is not a part of.
In the python repository changes have been made and a pull request opened (python/cpython#214) so that `http.cookies` could eventually support "SameSite" natively in 3.7.

However, since this is future stuff it doesn't really make a difference for bottle at the moment and i've come to the conclusion that it would be feature that's nice to have.

So i've added this feature to bottle, and updated the docs.

It doesn't to much, i just import the `_reserved` dict in which valid attributes for cookies are stored,
`set_cookie()` now supports the keyword `same_site` and ` raises a `CookieError` if the value for `same_site` isn't `strict` or `lax`.

This mimics petty much what the requested change for `http.cookies` in 3.7 would do.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -128,7 +128,7 @@ if py3k:
     from urllib.parse import urljoin, SplitResult as UrlSplitResult
     from urllib.parse import urlencode, quote as urlquote, unquote as urlunquote
     urlunquote = functools.partial(urlunquote, encoding='latin1')
-    from http.cookies import SimpleCookie
+    from http.cookies import SimpleCookie, Morsel, CookieError
     from collections import MutableMapping as DictMixin
     import pickle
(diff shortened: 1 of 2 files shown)


defnull:
This is a reasonable workaround. Thanks :)

We just have to remember adding a feature-gate as soon as python supports this natively.

provinzkraut:
It was my pleasure.. (=

I'm watching the open pull request anyway, so i'll have that in mind if the time has come.

PR #986, opened 2017-07-18
bugfix: import Morser for python 2

presveva:


defnull:
Oh I missed that. Thanks :)

PR #987, opened 2017-07-22
Fix CI 

oz123:
@defnull I removed Python 3.2 version because they break the build. 
Also, this version of Python is obsolete and it should cause people to think bottle in it's
current master branch is not working with their current Python (I assume most people 
will not have Python 3.2).


defnull:
I'm testing 3.2 in a non-public CI anyway, so dropping the tests from public CI is totally fine. Thanks :)

PR #1013, opened 2017-11-29
Add UnicodeFormsPlugin + tests

### LINKED ISSUE #774: Formsdict attribute like access returns invalid values for parameters submitted with unicode characters
Suppose we got a parameter submitted which has the value of 'ğüşiöç': `aparam = 'ğüşiöç'`

If request.method is 'GET':

`request.GET.aparam returns : 'ğüşiöç'` --> ok
`request.GET['aparam'] returns : 'ÄÃ¼ÅiÃ¶Ã§'`--> ??

If request method is 'POST':

`request.POST.aparam returns : ''` --> ??
`request.POST['aparam'] returns : 'ğüşiöç`--> ok

This behaviour interrupts clean and consistent coding style. When I need to access request form parameters I need to remember correct way to access them rather than simply writing request.POST.aparam.

Is this behaaviour is intended way or is this a bug or am I missing sth?

PS: Tested with Python 3.4, Python 3.2, and latest bottle github clone code

TIA

logannc:
Fixes #774 .

What: Adds a built-in plugin, `UnicodeFormsPlugin`, which looks for `config['utf8.unicode_forms']`. By default, `utf8.unicode_forms` is `False`, which means `FormsDict` will be used. When it is set to `True`, `UnicodeFormsDict` is used which inherits from `FormsDict` and overrides `MultiDict` methods so that all methods use `_fix` when applicable. It still retains access to the raw values through a few methods.

Usage:
```
app = bottle.Bottle()
app.install(bottle.UnicodeFormsPlugin())

app.config['utf8.unicode_forms'] = True
# or
@app.post('/someroute', **{'utf8.unicode_forms' : True})
def myroute():
    pass
```

It is both skippable and configurable per-route.

Since #774 is a compatibility issue, I tried to make sure an easy migration was possible. People can start using this plugin now, it can be turned on by default at a later version, and then we can remove the plugin and change the default behaviour later.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1204,13 +1204,17 @@ class BaseRequest(object):
     def get_header(self, name, default=None):
         """ Return the value of a request header, or a given default value. """
         return self.headers.get(name, default)
+    
+    @property
+    def form_container_type(self):
+      return UnicodeFormsDict if 'bottle.route' in self and self.route.config.get('__utf8.use_unicode_forms_dict') else FormsDict
 
     @DictProperty('environ', 'bottle.request.cookies', read_only=True)
     def cookies(self):
         """ Cookies parsed into a :class:`FormsDict`. Signed cookies are NOT
             decoded. Use :meth:`get_cookie` if you expect signed cookies. """
         cookies = SimpleCookie(self.environ.get('HTTP_COOKIE', '')).values()
-        return FormsDict((c.key, c.value) for c in cookies)
+        return self.form_container_type((c.key, c.value) for c in cookies)
 
     def get_cookie(self, key, default=None, secret=None, digestmod=hashlib.sha256):
         """ Return the content of a cookie. To read a `Signed Cookie`, the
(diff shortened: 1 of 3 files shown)


logannc:
Hm. I'm pretty sure that failing test is not related to my changes. Possibly needs `group: deprecated-2017Q3` per https://github.com/travis-ci/travis-ci/issues/8315 and related issues?

logannc:
@defnull Do you have time to take a look at this? I also notice that some other PRs tests are now passing, but I don't seem to be able to restart the travis build for my PR.

airking05:
It seems that there is no wrong part about travis but an error occurs...

logannc:
@airking05 The current error is in installing a package through pip. Pretty sure I didn't break pip.

logannc:
@airking05 See? Fluke. :)

@defnull Any thoughts?

logannc:
I have discovered a flaw in this API. In an attempt to be backwards compatible, I default to using `FormsDict` when a route config is not available. This means that accessing `params` or any other `FormsDict` based property on the `bottle.request` object during a `before_request` hook will lock in using `FormsDict` for the accessed property.

logannc:
Now it defaults to the app config, then the route config. If you want to override it at the route level... you have to make sure that your `before_request`s don't touch a property you want to use the route's config. 

We *could* do more by wrapping the route callback in a wrapper which could inspect the properties and convert them, as needed, but that is a lot of work for the very niche case of the app and route configs disagreeing and not being able to use `raw=True` when querying the `UnicodeFormsDict`.

logannc:
I've also removed the `UnicodeFormsPlugin` whose original purpose was to act as a buffer between the app level configuration and the actual code. Since we are now directly using app level configuration, it would have been more than useless.

logannc:
:telescope: Anyone here?

PR #1014, opened 2017-11-29
A few comment cleanups

nzufelt:
Several lines had misaligned `#` symbols in them, they were:
271, 618, 1147, 1965, 2069, 2790, 2926, 3183, 3578, 3768, and 4260.


defnull:
The additional `#` was intentional :) I think it looks nice

PR #1018, opened 2017-12-09
fix pass None to tob

Hanaasagi:
I think the `tob` fuction has a bug on this [line](https://github.com/bottlepy/bottle/blob/master/bottle.py#L164)

```Python
def tob(s, enc='utf8'):
    if isinstance(s, unicode):
        return s.encode(enc)
    return bytes("" if s is None else s)
```

If pass `None` to `tob`, it will execute `bytes("")`. And there will be a `TypeError` that `string argument without an encoding`

I think you mean that `return b'' if s is None else bytes(s)`

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -161,7 +161,7 @@ else:  # 2.x
 def tob(s, enc='utf8'):
     if isinstance(s, unicode):
         return s.encode(enc)
-    return bytes("" if s is None else s)
+    return b'' if s is None else bytes(s)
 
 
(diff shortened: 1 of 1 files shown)


defnull:
Yes this is a but, but the fix won'd work with python 2 as `str(None)` returns "None" as a string.

defnull:
Wait, we dropped Python 2.6 support did we? Great :D Forget what I was saying. Thanks for the PR

Issue #1027, opened 2017-12-30
Indentation in generated code is wrong following embedded block


vsajip:
The following template:
```
<%
def foo():
  return 'foo'
%>
<p>test</p>
<%
def bar():
  return 'bar'
%>
```
generates the following code:
```

def foo():
  return 'foo'
  
  _printlist((u'<p>test</p>\n',))
  
  def bar():
    return 'bar'
    

```
This doesn't seem right. The `_printlist` call and the `def bar():` shouldn't be indented.

Jwink3101:
I am not at a real computer to test but I think you are forgetting (or not reading the documentation) that says all indentations are ignored and you need to end the block with

    %end

Try that and see!

Jwink3101:
Just to follow up, there is no `%` for a block before the end. Just after any kind of block start, add an `end`

vsajip:
Yes, but this is inside an embedded block, so you can't use % end. Hence you can do
```
% if foo:
<p>Foo</p>
% end
```
but not inside an embedded multiline block delimited by
```
<%
...
%>
```
I _have_ read the documentation, but seeing that _some_ indentation is preserved inside multiline blocks, I'm trying to understand the limitations more precisely. Note that in my second example in the OP, the indentation  is actually correctly preserved, except for the last line.

defnull:
You always have to close blocks with `end`. If the documentation is not clear enough, I'd be happy to accept pull requests.

ralphwetzel:
To conclude: You can't use `% end` ... but you **can** use `end` inside an embedded block:

```
<%
def foo():
  return 'foo'
  end
%>
<p>test</p>
<%
def bar():
  return 'bar'
  end
%>
```

This returns:

```
def foo():
  return 'foo'
_printlist(('<p>test</p>\n',))
def bar():
  return 'bar'
```

defnull:
> I have read the documentation, but seeing that some indentation is preserved inside multiline blocks, I'm trying to understand the limitations more precisely. Note that in my second example in the OP, the indentation is actually correctly preserved, except for the last line.

The indentation is not preserved, but recreated by bottle while parsing the template. The template engine contains a very simple modified python parser. It detects the `if foo:` lines a block start and will indent everything up until the next `end` (with special rules for `else`, `expect` and other block continuation keywords).

Try this:

    <%
         def foo():
      return 'foo'
    end
    %>

This will produce valid python code just fine :)

defnull:
[...]

PR #1028, opened 2018-01-01
Updated documentation on SimpleTemplate and embedded blocks.

vsajip:
As discussed in #1027.

### DIFF
--- a/docs/stpl.rst
+++ b/docs/stpl.rst
@@ -95,6 +95,48 @@ Both the ``%`` and the ``<%`` tokens are only recognized if they are the first n
 
 If you find yourself needing to escape a lot, consider using :ref:`custom tokens <stpl-custom-tokens>`.
 
+The use of ``end`` tokens in embedded blocks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Contents of code blocks surrounded by ``<%`` and ``%>`` might need to be not actually valid Python as understood by the Python parser, but close to it. This is because the content of these blocks is not actually parsed by Python, but rather by the ``SimpleTemplate`` parser, which then translates it to valid Python code. Thus, as per the second additional syntax rule mentioned above, the ``end`` token (without a preceding ``%``) must be used to help the parser understand how to indent the generated code correctly. As an example, the template code::
+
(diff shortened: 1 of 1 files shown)


vsajip:
> This is more than needed.

I get this, but

> It also comes off a bit editorialized

not sure what you mean by this.

Jwink3101:
The not being valid python  it is valid except for the `end` statement. 

Also, on another note, simple template doesn’t evaluate the code. It just parses for the key words and then python evaluates that. It’s an important distinction because the engine isn’t reimplementing logic.

vsajip:
> The not being valid python it is valid except for the end statement.

Yes, I wrote

> might need to be not actually valid Python as understood by the Python parser, but close to it.

which would seem to cover it. I don't mind changing *close* to *very close*.

I also didn't say that `SimpleTemplate` evaluated the code - I used the terms *parsed*, *translates* and *generated*, which seem appropriate in this context.

defnull:
The examples are a little misleading I think. The intendation counter is not reset at the `%>` token. Not using `end` to close a python block is always wrong, even if the result might look right.

I prefer to reason about this in two steps: `%` and `<%%>` are used to mark single lines or multiple lines as code. All lines that are not code are translated into `_printline(...)` statements. *Then* the result is re-intended as if it were valid python code with an aditional required `end` keyword.

The key message is that `%` and `<%%>` work exactly the same way. The latter is only a convenient way to type less for longer code segments.

vsajip:
I get this now. The thing that threw me was that every example in the current documentation has `% end` on single lines, and there are no examples of using `end` in a `<% %>` block. I naïvely tried `% end` inside the `<% %>`, which of course didn't work, and that led to me raising #1027. Following the discussion there, I knew what I had to do; I only wanted to update the documentation to make it clearer for anyone else who ran into this.

Would it be better to have an example with *more* levels of indentation, and mixing `<% %>` and `%` lines and showing the generated result? Or is that overkill, as @Jwink3101 seems to think? Or perhaps just a very short update along the lines of

>  `%` and `<% %>` work exactly the same way. The latter is only a convenient way to type less for longer code segments.

I think it would be good to have an example of a `<% %>` block which shows indented blocks and `end`.

PR #1032, opened 2018-01-16
Fix UnboundLocalError in the ConfigDict load_dict method

an9wer:
The error is raised when I try the following code in the interpreter:
```
>>> from bottle import ConfigDict
>>> c = ConfigDict()
>>> c.load_dict(1)
  File "bottle.py", line 2114, in load_dict
    raise TypeError('Source is not a dict (r)' % type(key))
UnboundLocalError: local variable 'key' referenced before assignment
```

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2111,7 +2111,7 @@ class ConfigDict(dict):
         while stack:
             prefix, source = stack.pop()
             if not isinstance(source, dict):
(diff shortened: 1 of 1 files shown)


defnull:
Good catch. The failing travis-ci builds for 2.5 and 3.1 can be ignored. Thanks :)

PR #1033, opened 2018-02-05
Fixed ResourceWarning: unclosed file <_io.BufferedReader... in static…

### LINKED ISSUE #1000: ResourceWarning: unclosed file <_io.BufferedReader... in static_file()
We've recently got this error coming from `static_file()` function ([this line](https://github.com/bottlepy/bottle/blob/master/bottle.py#L2910)):

> /usr/lib/python3.5/unittest/case.py:600: ResourceWarning: unclosed file <_io.BufferedReader name='/tmp/20170926_084744_545761.pdf'>
  testMethod()

`with` should fix it as it did in a similar issue: https://github.com/bottlepy/bottle/pull/611. Thanks.

airking05:
Hello.
I have fixed #1000 issue.
I would like you to review :)

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2813,9 +2813,11 @@ def _file_iter_range(fp, offset, bytes, maxread=1024 * 1024):
     fp.seek(offset)
     while bytes > 0:
         part = fp.read(min(bytes, maxread))
-        if not part: break
+        if not part:
+            break
         bytes -= len(part)
         yield part
+    fp.close()
 
 
 def static_file(filename, root,


airking05:
@oz123 thank you for your reviewing! : )
I had the same idea that we should use with .. statement, 
but i realized that def static_file() should return file object.
In addition the file object is read outside.
So we cannot fix #1000 problem as with ... statement.

oz123:
@airking05 it seems you are correct. The tests are failing for my PR ...
Then, despite being a weired solution, I say it is a valid one.

airking05:
@oz123 I agree with you. This is a weired solution... 
And this solves only the case range_header is set and request.method is not 'HEAD'.
When range_header is not set, we may get the error 'unclosed file'...

I want to solve all the def static_file() but It seems difficult :cry:

defnull:
I cherry-picked this one. Thanks :)

defnull:
>  When range_header is not set, we may get the error 'unclosed file'...

Not really. WSGI calls 'close()' on the returned request body, if defined, after transmission. This did not work for the reange-header case because the file was hidden within an iterator. But it should work (and should always have worked) for non-range-requests.

airking05:
@defnull Thanks for your kind instruction. Now I grasp how bottle and wsgi work :+1:

PR #1034, opened 2018-02-13
Fix path_shift fall back always triggering

Voultapher:
Instead of always doing a negative root path shift,
even after a successful return, only catch exceptions.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -737,7 +737,8 @@ class Bottle(object):
                 body = app(request.environ, start_response)
                 rs.body = itertools.chain(rs.body, body) if rs.body else body
                 return rs
-            finally:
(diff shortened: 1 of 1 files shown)


Voultapher:
@defnull the last time this was changed, is 7 years ago by you. I fail to understand why doing `request.path_shift(-path_depth)` even after a successful return could be the wanted behavior. Can you please shine some light on this.

defnull:
If I remember correctly, because `after_request` handlers installed to the parent app should see the original state, not the modified one.

Voultapher:
Ok, let me see if I can come up with a test for it.

Voultapher:
The use case no longer matters for me so I'm closing this.

PR #1035, opened 2018-02-13
Alternative fix for #1000

oz123:
this fix is a bit more intrusive, but it improves the overwhole
 readability of the code (compared to #1033)


defnull:
This patch would cause large file downloads to be read into memory and not streamed to the client.

PR #1038, opened 2018-02-16
[Fix] incorrect Python syntax

airking05:
Hi.
I'd like you to review this fix.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -457,7 +457,7 @@ class Router(object):
 
     def build(self, _name, *anons, **query):
         """ Build an URL by filling the wildcards in a rule. """
-        builder = self.builder.get(_name)
+        builder = self.builder.get(_name, )
         if not builder:
             raise RouteBuildError("No route with that name.", _name)
         try:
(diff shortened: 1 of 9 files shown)


defnull:
This does not fix anything. It does not even improve readability.

bottle.py
```diff
@@ -2246,9 +2244,9 @@ def replace(self, key, value):
         self.dict[_hkey(key)] = [_hval(value)]
 
     def getall(self, key):
-        return self.dict.get(_hkey(key)) or []
+        return self.dict.get(_hkey(key), ) or []
 
-    def get(self, key, default=None, index=-1):
+    def get(self, key, default=None, index=-1, **kwargs):
```
airking05: I mean I wanted to change this get() args matched with MultiDict.get() args.

test/test_server.py
```diff
@@ -66,13 +66,13 @@ def setUp(self):
     def tearDown(self):
         if self.skip: return
 
-        if self.p.poll() == None:
+        if self.p.poll() is None:
```
airking05: This PR includes above changes about None comparison.
I think it would be better to use 'is' not '=='.

bottle.py
defnull: Ahh okay, I did not see that between all the tailing-comma-edits all over the place.

But it is not passed along to MultiDict.get() and not used, so why include it in the public API?

test/test_server.py
defnull: While you are right (one should use `is` for `None` checks) in this case it does not matter much. The current implementation is not wrong.

defnull:
Main reason for not pulling this are the ~70 added tailing comma edits. This is just noise. You should try to configure your editor or tool in a way that those are not added automatically.

PR #1044, opened 2018-02-28
Fix resources leak in tests

bmwant:
## Rationale
We have a lot of `ResourceWarning: unclosed file <_io.BufferedReader name` due to some files being improperly closed in tests.
This pr ensures that resources is closed after they have been used.

## Changes
* Close opened file descriptors
* Some codestyle fixes

### DIFF
--- a/test/test_app.py
+++ b/test/test_app.py
@@ -7,6 +7,7 @@
 import unittest
 from bottle import Bottle
 
+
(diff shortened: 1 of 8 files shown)


bmwant:
@defnull could you please review?

bmwant:
Looks like this one is not needed, so closing for now

PR #1068, opened 2018-05-30
Fix for proper decoding non-latin1 symbols

vyt777:
If there are cyrillic or other non-latin1 symbols in data in request and there are simultaneously files in request there will be UnicodeEncodeError because these symbols cannot be encoded in latin1. These error appears only if there are files in request, e.g. content-type header is multipart/form-data. If there are no files in request and content-type header is application/x-www-form-urlencoded then will be no UnicodeEncodeError (cyrillic symbols in data decoded properly).

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1398,7 +1398,7 @@ class BaseRequest(object):
             for key, value in pairs:
                 post[key] = value
             return post
-
+        
         safe_env = {'QUERY_STRING': ''}  # Build a safe environment for cgi
         for key in ('REQUEST_METHOD', 'CONTENT_TYPE', 'CONTENT_LENGTH'):
             if key in self.environ: safe_env[key] = self.environ[key]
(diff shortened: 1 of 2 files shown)


defnull:
This unfortunately only fixes the symptom, not the problem. The default encoding for multipart in Python 3 FieldStorage is utf-8 and does not need re-encoding, but everything else is encoded latin1 according to PEP-3333 and needs re-encoding. `FormsDict.recode_unicode` should be set to false for multipart POST data in Python 3. best place to do that is probably `Bottle.POST()`.

vyt777:
Yes, that seems right to me. Don't know should I start a new pull request with this or just add another commit, so I edited my pull request for this fix in commit #1068 .

defnull:
Same pull request is fine. Could you add a small test that fails for the old behavior to prevent regressions? Or I'll add the test this weekend and pull.

vyt777:
Yes, I would love to add a small test

PR #1071, opened 2018-06-10
Flake8 finds Python syntax errors and undefined names

cclauss:
[flake8](http://flake8.pycqa.org)

These tests should pass ~once #997 is merged.~


defnull:
Could you move the `pip install flake8` into `test/travis_setup.sh`? Pip sometimes fails or times out. The retry helps to avoid false failures.

cclauss:
Tracks with https://devguide.python.org/#branchstatus

PR #1079, opened 2018-07-19
Travis CI: remove Python 3.3 and add Python 2.7

cclauss:
* Python 3.3 is went EOL last year  https://devguide.python.org/#branchstatus
* Python 3.7 is shipping  https://github.com/travis-ci/travis-ci/issues/9069


defnull:
Good point, thanks

PR #1096, opened 2018-09-27
Provide exc_info to start_response() when handling exceptions - 0.12

braedon:

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -874,6 +874,7 @@ class Bottle(object):
             if not self.catchall: raise
             stacktrace = format_exc()
             environ['wsgi.errors'].write(stacktrace)
+            environ['bottle.exc_info'] = sys.exc_info()
             return HTTPError(500, "Internal Server Error", _e(), stacktrace)
 
     def _cast(self, out, peek=None):
@@ -950,6 +951,7 @@ class Bottle(object):
 
     def wsgi(self, environ, start_response):
         """ The bottle WSGI-interface. """
+        provided_exc_info = False
         try:
             out = self._cast(self._handle(environ))
             # rfc2616 section 4.3
@@ -957,11 +959,16 @@ class Bottle(object):
             or environ['REQUEST_METHOD'] == 'HEAD':
                 if hasattr(out, 'close'): out.close()
                 out = []
-            start_response(response._status_line, response.headerlist)
+            exc_info = environ.get('bottle.exc_info')
+            if exc_info is not None:
+                del environ['bottle.exc_info']
+                provided_exc_info = True
+            start_response(response._status_line, response.headerlist, exc_info)
             return out
         except (KeyboardInterrupt, SystemExit, MemoryError):
             raise
         except Exception:
+            if provided_exc_info: raise
             if not self.catchall: raise
             err = '<h1>Critical error while processing request: %s</h1>' \
                   % html_escape(environ.get('PATH_INFO', '/'))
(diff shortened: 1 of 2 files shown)


braedon:
Hi @defnull, 

What was the reason for closing this PR?

The current exception handling doesn't allow middleware to see exceptions, making it hard to implement really useful things like structured logging. It'd be great if this, or some other solution, could be merged!

Thanks,
Braedon

defnull:
I usually do not close pull requests without an explanation. Is there a matching issue? This was closed a year ago and I do not recall the exact reason. But I can imagine some:

* This adds logic to the hot path that is only needed in some very narrow corner case: Handling of exceptions in middleware, even if bottle is configured to catch and handle all exceptions (catchall=True). I'd say this is kind of unusual.
* If you want full control over exceptions, you can set `app.catchall=False` and Bottle will no longer catch anything but its own exceptions. This is the go-to approach for the typical in-browser debugger middleware.
* You could also install an error handler or plugin that does the logging. Middleware is a very low-level and error-prone approach, and should usually be avoided if there are more high-level hooks available.

braedon:
Thanks for reply. Sorry I didn't bring this up a year ago - I must have missed the email notification. I can't remember/find a matching issue, though there is this PR against 0.13: https://github.com/bottlepy/bottle/pull/1097

I understand the reluctance to merge this kind of change, but if I can try and make a case for it...

I'll grant you this may be somewhat of a corner case, but as far as I can tell (I may be reading it wrong) it is still a requirement in the WSGI spec. Specifically, the referenced error handling section says that application authors should:

> Always provide exc_info when beginning an error response

In this case we know `start_response()` hasn't been called before, so `exc_info` isn't needed to allow `start_response()` to abort if data has already been sent, but it's still the beginning of an error response none the less. The section talks specifically about middleware relying on this parameter.

In terms of implementations that don't use middleware, I wasn't able to find a reasonable alternative to do what I need. Specifically, I'm using middleware to provide structured request logging, including exceptions, replacing the inbuilt plain text logging in bottle and wsgi servers.

I can't remember all the details - it's been too long - but I started out trying to use a plugin, but couldn't get it working reliably. Since plugins run before Bottle has finished fully processing a response (e.g. cast values, added certain headers, handled exceptions) it's difficult to get all the values needed for a reasonable request log.

Logging exceptions could be done in an error handler, but that would then be separate from the main request logging, making the solution more complicated to install. Any other use of error handlers would also conflict, making things more complicated still.

Finally, `app.catchall=False` isn't really what I'm after. I'm not looking to replace Bottle's error handling - I very much want it to keep doing what its doing - I just want to log exceptions as they go by.

bottle.py
```diff
@@ -950,18 +951,24 @@ def _cast(self, out, peek=None):
...
+            exc_info = environ.get('bottle.exc_info')
+            if exc_info is not None:
+                del environ['bottle.exc_info']
+                provided_exc_info = True
+            start_response(response._status_line, response.headerlist, exc_info)
             return out
         except (KeyboardInterrupt, SystemExit, MemoryError):
             raise
         except Exception:
+            if provided_exc_info: raise
```
braedon: When reviewing this again I noticed that the above `start_response` is never called multiple times (though the below one might be). 

I included this change because of the following line in the [error handling section](https://www.python.org/dev/peps/pep-3333/#error-handling) of pep-3333:

> Never trap errors raised by start_response when exc_info is being provided

However, if I'm reading the intent behind it correctly, it's not actually needed here. Even though we're adding `exc_info` to the above `start_response()`, the server can't have sent any data yet (it's our first call), so `start_response()` will never need to raise the type of exception we're not allowed to trap.

Removing this line (and others related to `provided_exc_info`) would avoid the possibility of edge cases where we passed `exc_info` to `start_response()` and then it raised an unrelated exception, which we then didn't catch.

PR #1097, opened 2018-09-27
Provide exc_info to start_response() when handling exceptions - 0.13

braedon:


braedon:
Superseded by https://github.com/bottlepy/bottle/pull/1221

Issue #1106, opened 2018-10-20
sameSite cookie attribute fails when using redirect


SlIdE42:
When setting a cookie and therefore using redirect, `sameSite` security attribute is not set but a falsy second cookie is created instead (with name `sameSite` and value `strict` or `lax`)... Even worse, if you use `secure` attribute, it's put on the wrong one.

You can reproduce it with the code below : 
1. Clean all cookies from localhost
2. Go to http://localhost:9000/ (meta redirect)
3. See one good cookie with `sameSite` attribute
4. Clean all cookies from localhost
5. Go to http://localhost:9000/?redirect (HTTP redirect)
6. See two bads cookies without `sameSite` attribute...

It has been tested with the latest development version of Bottle, Firefox 62 and Chrome 69, on Linux and macOS.

By the way, thank you for this small but useful piece of code !

```python3
#!/usr/bin/python3

import bottle

@bottle.route('/')
def index():
    bottle.response.set_cookie('name', 'value', same_site='strict')

    if bottle.request.query.get('redirect') == None:
        return """<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Redirect</title>
    <meta http-equiv="refresh" content="0; URL=/cookie">
  </head>
  <body>
  </body>
</html>"""
    else:
        bottle.redirect('/cookie')

@bottle.route('/cookie')
def redirect():
    bottle.response.content_type = 'text/plain; charset=UTF8'
    return """Please clean all cookies before each run !

Cookies :
=========
{}

Good value : "name=value" (with sameSite set to strict)
Bad  value : "name=value; sameSite=strict" (two cookies without sameSite)
""".format(bottle.request.headers.get('Cookie'))

@bottle.route('/favicon.ico')
def favicon():
    bottle.response.status = 204
    return ""

if __name__ == '__main__':
    bottle.run(host='localhost', port=9000, debug=True)
```

defnull:
If that is the case, then the way bottle copies cookies from one response instance to the next is flawed:  https://github.com/bottlepy/bottle/blob/master/bottle.py#L1683

SlIdE42:
This could be easily fixed by replacing `same_site` and `same-site` by `samesite` (pull request #1107)

PR #1107, opened 2018-10-20
Fix issue #1106 with sameSite cookie attribute

SlIdE42:
Fix issue #1106 with sameSite cookie attribute : 

When setting a cookie and therefore using redirect, `sameSite` security attribute is not set but a falsy second cookie is created instead (with name `sameSite` and value `strict` or `lax`)... Even worse, if you use `secure` attribute, it's put on the wrong one.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1815,7 +1815,7 @@ class BaseResponse(object):
             :param secure: limit the cookie to HTTPS connections (default: off).
             :param httponly: prevents client-side javascript to read this cookie
               (default: off, requires Python 2.6 or newer).
-            :param same_site: disables third-party use for a cookie.
+            :param samesite: disables third-party use for a cookie.
               Allowed attributes: `lax` and `strict`.
               In strict mode the cookie will never be sent.
               In lax mode the cookie is only sent with a top-level GET request.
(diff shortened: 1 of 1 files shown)


SlIdE42:
Thank you for trying it but are you really sure ? I just check again and it seems to work with Chrome 70 and Firefox 63 (please note that parameter `same_site` has been changed to `samesite` within `set_cookie` function)...

defnull:
Yes and no. Your pull request actually fixes the issue, but at the wrong location ;) The bug is not in `set_cookie`, but in `BaseResponse.copy()`.

The `set_cookie` method works as expected, when using `same_site=strict` as documented. The `Morsel._reserved` dict is responsible for translating the snake-case attributes to their correct name.

The `BaseResponse.copy()` method, however, tried to serialize and then deserialize `SimpleCookie` to make a deep copy. This did not work for the `same-site/SameSite` special case, because the parser checks the keys of the `Morsel._reserved` dict, not the value, which it should. This may actually be a bug in the `Morsel` implementation in the stdlib, but one that only surfaces if you add your own keys to the private `Morsel._reserved` dict, and will probably not be fixed.

The more robust way to deep-copy a `Morsel` instance and its attributes is to use the `Morsel.update` method inherited from `dict`, I think Also, I really do not want to break backwards compatibility by renaming a `set_cookie()` parameter.

Thanks anyway :) I'll push a fix based on your report, soon.

defnull:
Hmm I think you were actually right, sort of. I fixed it in both locations now.

PR #928, opened 2017-01-19
bugfix: support multiple proxies ('HOSTNAME, HOSTNAME2' syntax)

danysirota:
take the first one in urlparts

PR #1109, opened 2018-10-29
Fix X-Forwarded-Host when multiple proxies, see PR#928

JonathanHuot:
See initial bugreport in the PR https://github.com/bottlepy/bottle/pull/928

This should fix the case where bottle is behind multiple reverse-proxy, and each proxy is adding its own X-Forwarded-Host.
When the request arrives at the bottle level, we have multiple values in `X-Forwarded-Host: www.foo.com, xyz.foo.local`. It makes any uses of `request.url`/`bottle functions` not possible. 

E.g. 
With `X-Forwarded-Host: www.foo.com, xyz.foo.local` given:

Before the patch:
`request.url == 'http://www.foo.com, xyz.foo.local/index.html'`

After the patch:
`request.url == 'http://www.foo.com/index.html'`

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1434,7 +1434,10 @@ class BaseRequest(object):
         env = self.environ
         http = env.get('HTTP_X_FORWARDED_PROTO') \
              or env.get('wsgi.url_scheme', 'http')
-        host = env.get('HTTP_X_FORWARDED_HOST') or env.get('HTTP_HOST')
+        if env.get('HTTP_X_FORWARDED_HOST'):
+            host = env.get('HTTP_X_FORWARDED_HOST').split(",")[0].strip()
+        else:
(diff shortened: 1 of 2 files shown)


JonathanHuot:
Any news on this ? Travis says the build is "pending" but it has been successfully completed.

defnull:
I'm still convinced that a multi-valued `X-Forwarded-Host` header is a configuration error in the reverse-proxy or load-balancer setup, and does not make any sense. The header should contain the original `Host` as sent by the client, so only the first public-facing proxy should set this header and any additional back-end hops should completely ignore it.

Note that `X-Forwarded-Host` is different from `X-Forwarded-For`, wihc *is* multi-valued and already recognized as such by bottle.

Issue #664, opened 2014-09-25
Add possibility to add headers to all outgoing error responses


holbech:
It would be super-helpful if Bottle allowed me to add headers on all outgoing responses - including 404s. 

I can do this on normal 200-responses using the after_request hook, as the response is ready in bottle.response when these are called, but when this hook is called on 404-error requests, the response is not ready yet (at least its status code is not 404), and any headers that I try to set are apparently dropped.

My use case is a REST endpoint that serves ajax calls from browsers, and I need to set the Access-Control-Allow-Origin header. I have done so according to the recipe at http://bottlepy.org/docs/dev/recipes.html#using-the-hooks-plugin , but it does not work for 404s and 500s. The consequence is that the consuming web app has no way of detecting what has gone wrong.

arulrajnet:
This should work for all response codes

```
class CustomHeaderMiddleware(object):

  def __init__(self, app):
    self.app = app

  def __call__(self, environ, start_response):
    def custom_start_response(status, headers, exc_info=None):
      headers.append(('Custom-Header', 'Some value'))
      return start_response(status, headers, exc_info)

    return self.app(environ, custom_start_response)
```

Then

```
from bottle import Bottle, run

app = Bottle()
app = CustomHeaderMiddleware(app)
run(app=app)
```

Now the curl output will be

```
arul@local ~$curl -v http://localhost:8080/
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.47.0
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 404 Not Found
< Date: Mon, 20 May 2019 17:13:59 GMT
< Server: WSGIServer/0.1 Python/2.7.12
< Content-Length: 720
< Content-Type: text/html; charset=UTF-8
< Custom-Header: Some value
< 
```

defnull:
The `after_request` hooks are called after 404 or other errors for a while now. I'm currently back-porting that to 0.12 so this should fix this issue.

PR #1110, opened 2018-11-02
Added hook for after response fully determined

tommilligan:
When using bottle, there is no hook for after the response is fully determined. This would be useful to me for logging the resulting status code of each response, after error handling (except critical handling) has occurred.

This PR adds an `after_response` hook.


cristicbz:
This seems related to #664

defnull:
You can do the same with WSGI middleware quite easily. I do not think we need another hook at that late state of the request handling. It'd not worth the additional error-handling, testing and documentation required, if WSGI middleware is already well known and documented.

Issue #995, opened 2017-09-07
(v0.13-dev) Can't run aiohttp server


Michael-F-Ellis:
```
Traceback (most recent call last):
  File "/Users/mellis/ls-mapped-mode/bottle.py", line 3709, in run
    server.run(app)
  File "/Users/mellis/ls-mapped-mode/bottle.py", line 3511, in run
    from aiohttp.wsgi import WSGIServerHttpProtocol
ImportError: No module named 'aiohttp.wsgi'
Bottle v0.13-dev server starting up (using AiohttpServer())...
```
Apparently aiohttp has obsoleted .wsgi. See https://github.com/aio-libs/aiohttp/issues/1108

Truth be told, my real interest here is to be able to run some background tasks as coroutines without having to resort to subprocesses or threads.  I read your writeup about using gevent but prefer to use asyncio and async/await since these are now built into Python.  Any suggestions?

Thanks!

hartwork:
For the record you need `pip install aiohttp<2` to make that work.

Related commit:
https://github.com/aio-libs/aiohttp/commit/9218c264d57c633bc21cb7a14cd7890a272d4e34

hartwork:
Related: #1118

defnull:
Removing a non-working adapter (e.g. rocket) is fine. Removing a somewhat-working adapter that might be used by some users may be problematic without a deprecation round. On the other hand, it is very easy to just copy&paste the <10 lines from the old bottle release to your own application to re-introduce a removed adapter. If upstream dropped support already, it should be fine if we do so, too, without a deprecation-round.

hartwork:
For `v0.13-dev` on `master` backend `aiohttp` uses `aiohttp-wsgi` as a bridge now supports `aiohttp>=2` (see 3d9a8c1d862a5fbbeb2676c283f5c42917702aa5). I believe this bug is fixed and can be closed.

defnull:
Rocket was dropped in master and retrospectively deprecated in 0.12, so we can close this I think. Thanks :)

hartwork:
Was that^^ comment was meant to go into #1119?

PR #1118, opened 2019-01-01
[waiting for review] Complete backend testing

hartwork:
Ready for first round of review / discussion :smiley: 

Dropping rocket is for #1119 
aiohttp fixes are for #995

### DIFF
--- a/.github/workflows/install-deps.sh
+++ b/.github/workflows/install-deps.sh
@@ -10,6 +10,6 @@ sudo apt-get install -y libev-dev
 pip install mako jinja2
 
 for name in waitress "cherrypy<9" cheroot paste tornado twisted diesel meinheld\
-            gunicorn eventlet flup fapws3 bjoern gevent uvloop; do
+            gunicorn eventlet flup fapws3 bjoern gevent "aiohttp<2" uvloop; do
     pip install $name || echo "Failed to install $name with $(python -V 2>&1)" 1>&2
 done
--- a/.github/workflows/install-deps.sh
+++ b/.github/workflows/install-deps.sh
@@ -10,6 +10,6 @@ sudo apt-get install -y libev-dev
 pip install mako jinja2
 
 for name in waitress "cherrypy<9" cheroot paste tornado twisted diesel meinheld\
-            gunicorn eventlet flup fapws3 rocket bjoern gevent uvloop; do
+            gunicorn eventlet flup fapws3 bjoern gevent uvloop; do
     pip install $name || echo "Failed to install $name with $(python -V 2>&1)" 1>&2
 done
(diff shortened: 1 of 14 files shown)


hartwork:
Any thoughts?

hartwork:
Any news?

hartwork:
Any news?

hartwork:
Any news?

defnull:
Thanks for your effort, but I'm in the progress of moving to github actions. I do not think that investing more time into travis-ci or circle-ci is worth it. Could you have a look at c9dd6808 and see if your changes can be integrated with the new (very basic) pytest based setup? I changed the install script for test dependencies so that it just ignores and logs install errors. That's probably not the best way to do it, but I'm unsure what to do about third party projects not catching up with new python releases. These should produce a warning, not an error. There is sure a lot to improve with the bottle test setup.

hartwork:
@defnull I force-pushed a no-op rebase to master just now and it seems like GitHub actions doesn't trigger any builds for it. Could you check if pull request builds are activated?

hartwork:
PS: If Travis CI should not be made to work, can we delete `.travis.yml` and turn it off? The middle ground with enabled but broken and breaking with new changes seems like trouble to me. What do you think?

defnull:
The plan is to remove travis-ci and circle-ci once gh-actions are working properly. Pull requests should now also trigger actions.

hartwork:
Thanks! 

Before I clean up in the wrong direction: Are you okay with me…

- dropping rocket, and
- replacing aiohttp<2 by aiohttp-wsgi

…in general?

defnull:
Please deprecate them first. Bottle has a rather strict backwards compatibility policy (one minor version). Add a `dept(0, 13, "reason", "fix")` warning if used at runtime, similar to cherrypy->cheroot. We can remove them completely in 0.14 then. If possible, continue testing with the last release that still worked. Dropping tests for completely broken dependencies (rocket) is fine, though. Broken things do not need to be backwards compatible.

hartwork:
After a bit of research I find that:
- Support for aiohttp — added in 884ef30e — never went into any release, it's only on `master`. (It needs Python 3 and `aiohttp<2` to work at all.) Do we need deprecation for aiohttp if it was never released?
- The only version of rocket that installs at all is 1.2.8 and it only installs with Python 2. So rocket support is as dead as it could be: There is no old working version today. Do we need deprecation for rocket?

defnull:
If it was never released, then aiohttp can go. And since we will drop Python 2 support with 0.13, rocket can also go. Good points.

hartwork:
Ready for review now

hartwork:
Any news?

PR #1120, opened 2019-01-02
Be more consistent about supporting Python 3.4 to 3.7

hartwork:
What do you think?


hartwork:
Any thoughts?

defnull:
Thanks a lot for your PR. 

How about we define (and document) for the future that the master branch (and next release) will only support python versions that are not EOL? Old bottle releases are maintained, so nobody is left out, even on slow moving enterprise OS distributions.

Issue #240, opened 2011-10-18
Bottle server crashes on reload when code has syntax error


coredumperror:
In my attempts to get around this myself, I've discovered that this may simply be impossible, but it would be really nice if the bottle server (`bottle.run(reloader=True)`) wouldn't crash out when it reloads a file with a syntax error.  

I realize that this may be impossible to fix since it's the interpreter itself throwing the exception, but if the server could somehow catch this and keep itself alive until the next reload it would be much more convenient.  The main reason I use the dev server instead of Apache is that I don't want to restart the server for every code change I make.  But I find that the bottle server has to be restarted almost as often as Apache because of syntax errors (and other import-time exceptions) that crash the app.

Maybe there's a way to code my wsgi script to handle this myself, but I haven't found a working method.

PR #1134, opened 2019-03-07
Fix a problem of crashing when reloading.

sam-yusuke:
Hello,

Thank you for nice framework.
I made a small patch to the reloading feature.
Could you take a look at this PR and give me a opinion about this approach.

## The situation to use this feature
This patch allow user to prevent exiting server when reloading SyntaxError containing file.
This patch is slightly related to the problem mentioned in https://github.com/bottlepy/bottle/issues/240, but this patch solve the different problem of running server with `python run_server.py` not `python bottle.py run_server.py`.

## How to test
1. Save and run code above.
```
import bottle

app = bottle.Bottle()

@app.get('test')
def test():
    return 'test'

bottle.run(app=app, port=9100, host='0.0.0.0', reloader=True)
```
2. Make some syntax error to the file while running.

Thank you!

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3659,6 +3659,7 @@ def run(app=None,
     if NORUN: return
     if reloader and not os.environ.get('BOTTLE_CHILD'):
         import subprocess
+        import py_compile
         lockfile = None
         try:
             fd, lockfile = tempfile.mkstemp(prefix='bottle.', suffix='.lock')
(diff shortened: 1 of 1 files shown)


defnull:
This would not catch a syntax error in any imported module, unfortunately, and even for small projects you usually have more than one module you are working on.

How about a `restart_on_errors` feature that reloads on all exit statuses, not only `3`? We just have to make sure to not accidentally prevent an ordered shutdown (Ctrl-C).

sam-yusuke:
@defnull 
Thank you for checking this PR!
OK. I will try your approach and commit again.

PR #1154, opened 2019-08-15
Skip If-Modified-Since check when client sends If-None-Match

### LINKED ISSUE #1132: Last modified time should be skipped when etag check fails.
It is possible that a change to file/directory causes a etag change and no last modified time change. For example:
1. Create folder A.
2. Create folder B.
3. Visit folder B.
4. Rename folder B to C, and then A to B.
5. Visit folder B (previous folder A).

At step 5, the client sents HTTP_IF_MODIFIED_SINCE header with the last modified time of folder B in step 2. Since the last modified time of folder A is less than B, the server will send 304 back, which is an error since the folder is actually changed.

To fix this, we should skip checking for 304 if the client already sends an etag which does not return 304. Here's a patch to fix this:

    diff --git a/bottle.py b/bottle.py
    index 0b22f2e..9ac9ee5 100755
    --- a/bottle.py
    +++ b/bottle.py
    @@ -2914,11 +2914,12 @@ def static_file(filename, root,
             if check and check == etag:
                 return HTTPResponse(status=304, **headers)
     
    -    ims = getenv('HTTP_IF_MODIFIED_SINCE')
    -    if ims:
    -        ims = parse_date(ims.split(";")[0].strip())
    -    if ims is not None and ims >= int(stats.st_mtime):
    -        return HTTPResponse(status=304, **headers)
    +    if not (etag and check):
    +        ims = getenv('HTTP_IF_MODIFIED_SINCE')
    +        if ims:
    +            ims = parse_date(ims.split(";")[0].strip())
    +        if ims is not None and ims >= int(stats.st_mtime):
    +            return HTTPResponse(status=304, **headers)
     
         body = '' if request.method == 'HEAD' else open(filename, 'rb')

Hanaasagi:
Skip checking `If-Modified-Since` if client already sends an `If-None-Match`.
It will resolves #1132 .

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2908,17 +2908,20 @@ def static_file(filename, root,
                                    clen, filename)
         etag = hashlib.sha1(tob(etag)).hexdigest()
 
+    inm = getenv('HTTP_IF_NONE_MATCH')
+
     if etag:
         headers['ETag'] = etag
-        check = getenv('HTTP_IF_NONE_MATCH')
-        if check and check == etag:
+        if inm and inm == etag:
             return HTTPResponse(status=304, **headers)
 
-    ims = getenv('HTTP_IF_MODIFIED_SINCE')
-    if ims:
-        ims = parse_date(ims.split(";")[0].strip())
-    if ims is not None and ims >= int(stats.st_mtime):
-        return HTTPResponse(status=304, **headers)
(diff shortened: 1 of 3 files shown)


Hanaasagi:
CI failed, because it could not find Python 2.7.3.

Issue #839, opened 2016-03-01
Fail to retrieve Content-Length from the HTTP response header


arnikz:
I'm using bottle v0.12.9. For my web app I need to retrieve the Content-Length value. I tried both
approaches:
i) response.get_header('Content-Length') -> returned None
ii) response.content_length -> returned empty string ''

Error msg for the latter:

File "/Library/Python/2.7/site-packages/bottle.py", line 1416, in **get**
    return self.reader(value) if self.reader else value
ValueError: invalid literal for int() with base 10: ''

Thanks.

defnull:
The client-provided "Content-length" header is in **request**.content_length, not **response**.

arnikz:
Thanks. I need to log HTTP requests to a file (in Common Log Format) rather to a console but can't access Content-Length in the response header. **reques**t.content_length returns **-1** instead of **983** (as shown below)

log to console (debug mode) gives
127.0.0.1 - - [02/Mar/2016 13:46:32] "GET / HTTP/1.1" 200 **983**

defnull:
A ´´request.content_length´´ of -1 indicates that the client did not provide a "Content-Length" header.

arnikz:
OK, I think, we talk about different things here.  A client application (e.g. curl) making a request to my bottle-based web app returns the following HTTP response (header):

HTTP/1.0 200 OK
Date: Wed, 02 Mar 2016 15:40:43 GMT
Server: WSGIServer/0.1 Python/2.7.10
Content-Length: **983**
Content-Type: text/turtle
Allow: GET

In the console I get the corresponding access log. 
127.0.0.1 - - [02/Mar/2016 13:46:32] "GET / HTTP/1.1" 200 **983**

I would like to fetch **this** number using the API but neither response.content_length nor request.content_length helps. So what do you suggest to use?

defnull:
Ahh okay! Unfortunately, bottle computes the content-length of a response only if it is not provided by the application or any plugin or hook. This happens at a very late moment in the pipeline, even after the `after_response` hook. The only way to get this number without computing it yourself is at WSGI layer.

This might help: http://blog.philippklaus.de/2013/06/wsgi-request-logger-logging-http-requests-with-any-wsgi-web-application-like-flask-bottle-or-django/

defnull:
[...]

PR #1156, opened 2019-08-26
[bug] set default value for "Content-Length"

luozhaoyu:
Setting default value as -1 when "Content-Length" is empty. If not, currently it will raise a ValueError

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1787,7 +1787,7 @@ class BaseResponse(object):
         return out
 
     content_type = HeaderProperty('Content-Type')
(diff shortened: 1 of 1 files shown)


agalera:
related to: https://github.com/bottlepy/bottle/issues/839

luozhaoyu:
how can I merge this PR?

luozhaoyu:
rebuild

defnull:
The build fails because travis dropped python 2.7.3 support. The PR is fine otherwise. Thanks :)

PR #1161, opened 2019-09-07
Fix syntax in webtest login recipe

oz123:
Previously, the example yielded SyntaxError. Adding an equal sign
fixes the assertion syntax.


defnull:
Thanks

PR #1167, opened 2019-10-08
Fix unix sockets in gunicorn

agalera:
Because the host and port were concatenated, it generated an incorrect socket name (```example.sock:8080```)
In gunicorn, if it starts with ```unix:``` it will create a unix socket


agalera:
@defnull can you check please? :)

Issue #1174, opened 2019-10-24
Add SameSite=None support in set_cookie


yoch:
Hi,

The latest RFC [introduces](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-5.3.7) a third option 'None' for SameSite, which is currently not supported.
In fact, all values but 'lax' and 'strict' must be treated as 'none'.

Regards

defnull:
Isn't `SameSite=None` equivalent to not setting this attribute at all?

provinzkraut:
As far as I understand, since [no behaviour is defined](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-5.3.7.1) for the value `none`, it *is* the same as not setting it at all.

yoch:
Not in practice, because some browsers (like [chrome](https://www.chromestatus.com/feature/5088147346030592) or firefox) define [SameSite=Lax by default](https://tools.ietf.org/html/draft-west-cookie-incrementalism-00).

Its not difficult to implement, I'll send a PR.

BTW, it is possible tu have this feature enabled in 0.12 ?

PR #1175, opened 2019-10-24
Enable SameSite=None

yoch:
See the discussion in #1174


yoch:
In fact, maybe the best option is to not raise an exception at all inj any case, because this is not the bottle responsibility to check that.

defnull:
The error for wrong values is nice to have to prevent security-compromising typos. I solved it a little bit differently, but it should work now. Thanks!

PR #1189, opened 2019-12-19
gunicorn: switch to use BaseApplication (fixes #756)

### LINKED ISSUE #756: prevent run() function pass argument parser object to gunicorn server
I have an argument parser that have an `--env, -e` option which is different from gunicorn's `--env, -e`. When I call run with the following parameters;

``` python
kwargs = { 'workers': 4 }
bottle.run(app.wsgi,
           host='localhost',
           port=8080,
           debug=False,
           reloader=False,
           server='gunicorn',
           **kwargs)
```

The `run()` function sends gunicorn to my previously created ArgumentParser object unexpectedly and gives the following error;

```
Error: environment setting 'production' invalid
```

This error is coming from gunicorn's config module;
https://github.com/benoitc/gunicorn/blob/master/gunicorn/config.py#L187
How can I solve this problem?
NOTE: I don't want to change --env option to some other thing

rcoup:
Gunicorn configuration is now only pulled from what is supplied through Bottle, not (for example) re-parsed from the commandline. Follows Gunicorn's [documented approach for integrated applications](http://docs.gunicorn.org/en/stable/custom.html).

* Before: `./my-bottle-app.py -v` → launches gunicorn, which re-parses `sys.argv`, finds the `-v`, prints the gunicorn version, and quits. Likewise, any command line argument which wasn't valid for gunicorn would cause an `error: unrecognized arguments: --whatever` message.
* After: launches gunicorn

Requires gunicorn >19.0 (Jun 2014)
Fixes #756

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3459,7 +3459,7 @@ class GunicornServer(ServerAdapter):
     """ Untested. See http://gunicorn.org/configure.html for options. """
 
     def run(self, handler):
-        from gunicorn.app.base import Application
+        from gunicorn.app.base import BaseApplication
 
         if self.host.startswith("unix:"):
             config = {'bind': self.host}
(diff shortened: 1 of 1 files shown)


defnull:
Oh they fixed it? Great! Thanks a lot for this find.

PR #1193, opened 2020-01-01
Removing Python 2 support from 0.13

defnull:
This is the integration branch for python 2 removal. Please submit pull requests against this branch and discuss the entire process here.


defnull:
Since Bottle-0.12 will be the last release supporting Python 2 (2.5+ actually) and very old versions of Python 3 (3.2+), it will become an LTS release and will be maintained (security patches only) for a bit longer than usual. This allows us to drop support for old python versions without hurting existing users too much.

I'm thinking about skipping 0.13 and move forward to 0.14 in a single step to make it more clear that there are breaking changes this time. Spamming all 0.12/Py2 users with a deprecation warning is a bit too much I think. Some people can't update, even if they want to.

oz123:
@defnull  this is very welcome. According to your own roadmap the aren't many blocking issues for 0.13. but it's been so long since the first map was written I'm not sure it's even relevant.
It's a good sign for people to get a new release out. Please go ahead and do this ASAP.

oz123:
@defnull also, please remove the classification from setup.py
https://github.com/bottlepy/bottle/blob/8561ffbc2ed2b16b4f7f90520d148e37e251d107/setup.py#L34.

You should also drop python 3.4. it's no longer maintained.

stigok:
Also remember to remove unsupported targets from the root Makefile

defnull:
The release-0.12 branch has its own Makefile and test suite. We can remove any trace of 2.7 or 3.4 support from the master branch.

cclauss:
It has been a year...  Conflicts.

defnull:
Yes, it probably wasn't the best idea to start with this before 2.7 is actually EOL and expect that there are no conflicts.

SergBobrovsky:
All declarations like 
```
class A(object):
``` 
may be reduced to
```
class A:
```

cclauss:
@SergBobrovsky That is an optional change but can be automated via tools like https://github.com/asottile/pyupgrade

SergBobrovsky:
remove
```
from __future__ import print_function
```

cclauss:
Also automated by pyupgrade.  Give it a try to see the changes that it would make.

clavedeluna:
Hey @defnull what's the status on this work? I'm happy to contribute but I saw your comment elsewhere stating you'd prefer this be done in smaller pieces?

.github/workflows/run_tests.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+        python-version: [3.5, 3.6, 3.7, 3.8]
```
clavedeluna, line 10: at this point, according to https://endoflife.date/python, 3.5 and 3.6 can be dropped too, though we may want to drop that later on

PR #1197, opened 2020-01-06
Fix typos in documentation.

tirkarthi:
Fixes some typos in documentation. The typos are in po files too. Not sure if I need to change them manually or it's done as part of build process.


tirkarthi:
Few more typos mulripart, deprectated and resticts. I will add another commit to fix them.

PR #1200, opened 2020-01-16
Fixed an issue using lighttpd

bernd-wechner:
Sadly using uwsgi under lighttpd the PATH_INFO environment variable is not set. Technically a lighttpd bug in all likelihood, but I did find this single one line fix is a workaround that sees my bottle app run fine under lightpd and uwsgi.

I'm not really sure this is the right patch, and don't mind in the least learning otherwise. It works on my app which is totally simple in one folder and I simply noticed REQUEST_URI was / and PATH_INFO not set. I have a git feel this doesn't generalise well, but hope this PR prompts a think about a more general path forward when PATH_INFO is missing. It is reverenced in other parts of bottle with defaults anyhow and only this one line crashes for me. So I patched it locally for now.

I will pursue an issue with lighttpd regarding the missing PATH_INFO as well but think it wise that bottle be robust against a missing PATH_INFO all the same.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -985,7 +985,7 @@ class Bottle(object):
         return tob(template(ERROR_PAGE_TEMPLATE, e=res, template_settings=dict(name='__ERROR_PAGE_TEMPLATE')))
 
     def _handle(self, environ):
-        path = environ['bottle.raw_path'] = environ['PATH_INFO']
+        path = environ['bottle.raw_path'] = environ.get('PATH_INFO',environ.get('REQUEST_URI','/'))
         if py3k:
             environ['PATH_INFO'] = path.encode('latin1').decode('utf8', 'ignore')
 


defnull:
`PATH_INFO` is a required WSGI parameter, so this is clearly a bug somewhere in lighttpd/uwsgi, probably in the lighttpd scgi module. Unfortunately `REQUEST_URI` means something different, and would no longer work as expected if the application is not located at the root path. So, this only works for you, but not for everyone. If you really need this workaround and cannot configure lighttpd/uwsgi to behave correctly, then a simple wsgi middleware would be sufficient. No need to patch bottle for that.

bernd-wechner:
Thanks for the feedback. I filed: https://redmine.lighttpd.net/issues/3005 as well. And concur that it's a lighttpd/scgi bug. I'm content with the little patch at my end in interim and hope the lighttpd issue gets fixed or a similarly useful response leading to a deployment solution in this context (lighttpd/scgi/uwsgi).

PR #1210, opened 2020-03-04
Fix broken link to WSGI specification

sharkwouter:
I tried testing this by building it with the bottlepy.org repository, but I wasn't able to get any link to work in the html files generated by that. This link is currently broken on the live page, though, because the capitalization doesn't match.


sharkwouter:
This is a minor documentation fix, is nobody even going to look at this?

PR #1103, opened 2018-10-13
Improve wsgi error handling 013

BubaVV:
Closes issue #1097 . Tests added to improve coverage a little bit

PR #1221, opened 2020-04-17
Improve wsgi error handling 013

braedon:

Supersedes https://github.com/bottlepy/bottle/pull/1097, and includes the tests from @BubaVV in https://github.com/bottlepy/bottle/pull/1103


defnull:
I do not fully understand if this is the best solution, but I'll trust you on this and merge. It only affects the error path, so it should not affect normal working applications. And it has tests \o/ Thanks!

braedon:
Thanks @defnull, Happy New Year!

PR #1223, opened 2020-04-23
Add module-level merge function

kyos2000:
Add merge function to the module-level functions like route function and mount function. This PR will be easy to merge the routes into the default application.


defnull:
Merging usually only happens if an application grows big enough to justify splitting it up in multiple parts. In that case I'd rather have people move away from the module-level shortcuts and be more explicit about application context. This change would send the wrong message, I think. The module-level stuff is targeted at really really simple one-module applications.

kyos2000:
@defnull   Thank you for the feedback.

PR #1228, opened 2020-05-05
Add bottle tools to plugins doc page

theSage21:
I made something based on bottle that allows us to auto-fill function arguments using request data.

```python
## This function can reside anywhere in your codebase

@fill_args
def some_function(json_key):
    # json_key will automatically be injected into the function call
    # from the POST data supplied in bottle.request
    pass

## You can then call it like this:

some_function()
```

I'm not sure if bottle accepts these kinds of pull requests but I thought I'd give it a shot anyways. :shrug: Thanks for the amazing project.


defnull:
Links are fine :) I'm more picky about PRs that try to add features to bottle itself. Thanks.

PR #1231, opened 2020-05-09
reuse port in Bjoern adapter

rcarmo:
This affords a considerable saving in terms of system resources.


defnull:
Looks save to do. Thanks

PR #1233, opened 2020-05-24
mutable default value {} as argument (pylint W0102)

zsxoff:


Jwink3101:
This is a style thing but I think the first patch should be `if globals is None` instead of `if not globals`. The advantage of this is that if the user for some reason wants to take advantage of the mutable argument and passes an empty dict. 

On the second patch, I think it’s actually okay since the passed dict is _only_ used to update an independent dict. It’ll still flag on a lint check but in practice, will have no ill-affects.

PR #1237, opened 2020-06-26
fix for latest cherrypy

sailfish009:
works for latest cherrypy version, tested.


defnull:
So CherryPy removed the WSGI server API in version 9 and re-added it in version 18? Hmm. I think we should finally drop support for cherrypy 9+. The WSGI part of it was outsourced into a spparate module (cheroot), so you do not need the full cherrypy dependency anymore to use its wsgi implementation. Supporting both (cherrypy and cheroot) makes no sense to me.

PR #1239, opened 2020-08-13
Adding new cors plugin

alfonsocv12:
I use this plugin to enable cors on all the routes on my api's i thought it was a good feature to have on bottle 

This is the way that i enable it on the bottle app
Screen Shot 2020-08-13 at 10 33 28 AM

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2042,6 +2042,45 @@ class TemplatePlugin(object):
         else:
             return callback
 
+class CorsPlugin(object):
+
+    name = 'cors'
+    api = 2
+
+    def __init__(self, cors_origins="*"):
+        '''
+        Constructor function
+        '''
+        self.cors_origins = cors_origins
+        self._options_route()
+
+    def apply(self, fn, context):
+        def _enable_cors(*args, **kwargs):
+            # set CORS headers
+            response.headers['Access-Control-Allow-Origin'] = self.cors_origins
+            response.headers['Access-Control-Allow-Methods'] = '\
+                GET, POST, PUT, PATCH, OPTIONS, DELETE'
+            response.headers['Access-Control-Allow-Headers'] = '\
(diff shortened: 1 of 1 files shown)


oz123:
This is not the correct place for this plugin IMO. There should be a dedicated plugin repository and a link in the documentation.

alfonsocv12:
Actually I was building the plugin library earlier but I saw like 3 different versions of the bottle-cors plugin no body solvent the problem this elegant, and with the implementation of core on all chromium browsers I believe its a must have event on a micro framework like bottle.

Thats why I put it behind the template plugin, its not really a necessity but its a most have, but if the team don't want the plugin I build the library for it thanks anyways.

alfonsocv12:
I saw that the plugin wasn’t working with abort errors so i add some more lines and the installation process is this one

I still believe that the best way of doing this is by adding it to the base bottle repo, i think the cors are standardized on today’s standards and in most cases of api development you will need to activate cors

Screen Shot 2020-08-13 at 2 09 40 PM

defnull:
Plugins should not bundled with bottle, but distributed separately. If it's bundled, it is a core functionality and does not need to be a plugin. I just merged a PR that adds a cors plugin to the plugin-page in the docs. Thanks for your work!

mrdc:
@alfonsocv12 
> Thats why I put it behind the template plugin, its not really a necessity but its a most have, but if the team don't want the plugin I build the library for it thanks anyways.

I think it's better to give a user control which `Access-Control-Allow-Methods` and `Access-Control-Allow-Headers` he want's instead of providing default ones.
I use hooks for that: it's like 10 lines of code in the beginning - very similar approach that you have in your proposal:
```python
@hook('after_request')
def enable_cors():
    response.headers['Access-Control-Allow-Origin'] = '*' 
```
The same for `Methods`. And it gives full control.

PR #1245, opened 2020-09-16
Clean eol python versions

oz123:
Bottle currently declares support for EOL versions of Python.
Also, some supported Python versions aren't properly published.


defnull:
Time to move on, yes. Thanks

PR #1246, opened 2020-09-28
Don't mutate headers dict passed into static_file()

braedon:
Mutating the passed in dict can produce unexpected results if the caller re-uses the dict, e.g. uses the same set of headers for multiple files.


defnull:
Thanks, good catch!

PR #1247, opened 2020-09-28
Fix build by dropping support for fapws3 server

braedon:
Github actions currently hangs indefinitely while trying to install fapws3 on Python 2.7, breaking the build. (It just fails on 3.x due to a print statement in the setup.py.)

fapws3 only works with Python 2, which is no longer supported. The project appears to be unmaintained - the last commit was 3 years ago.

(Note: I spent quite a while trying to work out why installing it no longer works on Github actions, without any luck. This seemed like the next best course of action.)

### DIFF
--- a/.github/workflows/install-deps.sh
+++ b/.github/workflows/install-deps.sh
@@ -5,11 +5,10 @@ set -x
 # Test utilities
 pip install -U pip pytest coverage
 
-# Test dependencies (Server back-ends and template engines)
-sudo apt-get install -y libev-dev
+# Test dependencies (template engines)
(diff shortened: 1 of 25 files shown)


defnull:
Python 2.7 is still supported by bottle to some extend, but fapws3 should be dropped, yes. Could you make FapwsServer emit a helpful deprecation warning and add some info to the changelog, so people updating at least know what is going on? Changelog is not really up to date, but at least breaking changes should appear there.

defnull:
I deprecated fapws3 and removed links from the docs. We can completely remove it in the future.

PR #1252, opened 2020-11-11
run_tests.yml: Add Python 3.9 to the testing

cclauss:


defnull:
Why did you remove the `cherrypy<9` restriction? (And why did test pass without it? I thought that their API changed with 9+)

cclauss:
I removed `cherrypy<9` because cherrypy is on v18 and https://github.com/bottlepy/bottle/runs/1386872236#step:5:63 which was caused by Python 3.9 cleanup of obsolete clutter... https://bugs.python.org/issue37804

defnull:
Yes, but cherrypy >= 9 did not work with the server adapter because they out-sourced their WSGI implementation it to a separate module (cheroot) and broke their API to use it from the cherrypy module. That's why we dropped support for cherrypy 9+ and users should move to cheroot instead.

defnull:
I'm wondering why the tests pass with cherrypy 18 again.

defnull:
I added newer python verions to test without the cherrypy change. Still think we should drop cherrypy in favour of cheroot. Thanks anyway!

cclauss:
Look like you will have to modify the tests or drop ancient 🍒py or hold off on Py39...
`AttributeError: 'WorkerThread' object has no attribute 'isAlive'`

Issue #1253, opened 2020-11-12
new server adapter


Ksengine:
I like to add new server adapter for [@Ksengine/Servelight](https://github.com/Ksengine/Servelight)
This server is based on wsgiref sourece code.
but with following updates.
- PEP 3333 for python 2
- python 2 and 3 both supported
- unix sockets supported
- Fix for IPv6 addresses
- SSL supported
- threading server - uses threads to handle requests. This is useful to handle web browsers pre-opening sockets, on which Server would wait indefinitely.
- multiprocessing server -handle each request in a new process.
- Prevent reverse DNS lookups
- Prevent ResourceWarning: unclosed socket on KeyboardInterrupt
- when given 0 as port use random port and update port to actual port

Ksengine:
i like to make a PR

Ksengine:
?

defnull:
See #1254

PR #1254, opened 2020-11-16
new server adapter

Ksengine:
**new server adapter for [@Ksengine/Servelight](https://github.com/Ksengine/Servelight)**
This server is based on wsgiref source code.
but with following updates.
- PEP 3333 for python 2
- python 2 and 3 both supported
- unix sockets supported
- Fix for IPv6 addresses
- SSL supported
- threading server - uses threads to handle requests. This is useful to handle web browsers pre-opening sockets, on which Server would wait indefinitely.
- multiprocessing server -handle each request in a new process.
- Prevent reverse DNS lookups
- Prevent ResourceWarning: unclosed socket on KeyboardInterrupt
- when given 0 as port use random port and update port to actual port

so I added server adapter
I had read -
- https://github.com/bottlepy/bottle/pull/647#issuecomment-60152870
- https://github.com/bottlepy/bottle/pull/865#issuecomment-242795341

according to #1253 Isuue

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3288,6 +3288,20 @@ class WSGIRefServer(ServerAdapter):
             self.srv.server_close()  # Prevent ResourceWarning: unclosed socket
             raise
 
+            
+class ServeLightServer(ServerAdapter):
+    def run(self, app):  # pragma: no cover
+        from sl.server import make_server
+        from sl.server import WSGIRequestHandler, ThreadingWSGIServer
+
+        handler_cls = self.options.get('handler_class', FixedHandler)
+        server_cls = self.options.get('server_class', ThreadingWSGIServer)
+
+        self.srv = make_server(self.host, self.port, app, server_cls,
+                               handler_cls)
+        self.port = self.srv.port  # update port actual port (0 means random)
+        self.srv.serve_forever()
+
 
 class CherryPyServer(ServerAdapter):
     def run(self, handler):  # pragma: no cover


Ksengine:
No reply

defnull:
No tests or docs, and it won't work I guess, since `FixedHandler` is not defined in that namespace.
As mentioned in #1255, there are too many WSGI implementations to support them all. New server adapters should have a clear benefit over setting up the server manually, more than just convenience. Thanks anyway.

Ksengine:
> No tests or docs, and it won't work I guess, since FixedHandler is not defined in that namespace.

no need to define `FixedHandler`. all bugs were fixed from @Ksengine/Servelight

Ksengine:
wsgiref adds PEP333 for python2 and PEP3333 for python3
@Ksengine/Servelight adds PEP3333 for any python version/implementation
this pr adds only few lines to to bottle

Ksengine:
I will make tests and docs

Ksengine:
would you like to accept

defnull:
A server adapter is not needed to use "Servelight" with bottle. As mentioned in #1255, there are too many WSGI implementations to support them all. New server adapters should have a clear benefit over setting up the server manually, more than just convenience.

Ksengine:
It has multi-threading and multi-processing servers

defnull:
I sense a communication problem here. Servelight may be great and have many benefits over wsgiref, but we are talking about adding a server adapter to bottle, not about how great Servelight is. You do not need a server adapter to use Servelight with bottle. It's actually harmful. A simple server adapter would hide most of the features Servelight offers. So why add it?

Ksengine:
Its featured but not popular yet.
bottle can use Its features.

defnull:
Server adapters are *not* a suitable way to promote new projects, sorry.

Ksengine:
> Server adapters are not a suitable way to promote new projects, sorry.

can I add description about ServeLight to bottle docs

Ksengine:
#1256

Jwink3101:
> can I add description about ServeLight to bottle doc

I am not in any position of power but I think this in a wikis may _eventually_ be the appropriate answer if ServeLight gains in popularity. 

But given its current status, my opinion is that it shouldn't be there! It's just too new without much popularity and, as you've said in your own replies, all it takes is two lines or so to use. So the right answer is to advertise in **your** repo how easy it is to use with Bottle. 

If it ever gains a real following _and_ has very clear advantages over what is there, then maybe it would make sense in a wiki-type page. But it doesn't need to go into the docs. 

I am not an admin or anything on Bottle but at this point, you're more-or-less spamming the issues. 

Just because you don't like the answer, it does not mean you should keep asking and posting. It is an inappropriate use of the Issues tracker and, at this point, counter productive.

Ksengine:
> all it takes is two lines or so to use

Hey, servelight is not a patch to wsgire
Its forked from wsgiref and improved
I had edited many lines to complete it.

2 lines for wsgiref means a patch

Ksengine:
Servelight works without wsgiref

defnull:
Read more carefully. We do not discuss Servelight here, but the proposed server adapter.

I do not know how to explain this more clearly: If a user wants to use Servelight with bottle, he or she can do that right now. No change needed. The documentation on how to use Servelight with a WSGI framework should be on the Servelight webpage, not here.

PR #1266, opened 2020-12-24
redundant brackets

SergBobrovsky:


defnull:
In some versions of Python, String.join() was actually significantly faster with a list than an iterator with unknown size, especially for small lists. May no longer be the case, though.

defnull:
Performance is not relevant in this case, I just wanted to share the reason why sometimes passing a list instead of a generator makes sense.

PR #1272, opened 2020-12-27
add _control_chars_in function

SergBobrovsky:


defnull:
```python3 -mtimeit -s 'key="Content-Type";import re;_control_chars_in = re.compile("[\n\r\0]+").search' '_control_chars_in(key)'
1000000 loops, best of 5: 253 nsec per loop
python3 -mtimeit -s 'key="Content-Type"' "'\n' in key or '\r' in key or '\0' in key"
5000000 loops, best of 5: 43.1 nsec per loop```

defnull:
Regular expressions are 5 times slower than three simple `in` checks and that check is run quire often during a request.

PR #1278, opened 2021-01-01
ResourceManager.__iter__()

SergBobrovsky:
a little shorter


defnull:
Are you sure that this does the same thing? Note that the search list is mutated within the loop.

SergBobrovsky:
Yes, it works right, I'm shure

PR #1282, opened 2021-01-01
refactor ConfigDict.load_module()

SergBobrovsky:
a little shorter


defnull:
That's also a lot less readable and harder to understand.

PR #1283, opened 2021-01-01
refactor WSGIHeaderDict._ekey()

SergBobrovsky:
a little shorter


defnull:
But also less readable. Single-line if-then-else should be used sparely.

PR #1284, opened 2021-01-01
refactor BaseRequest.script_name

SergBobrovsky:
f-string better and shorter
script_name as local var inside func script_name() - not good


defnull:
f-strings are not used anywhere in bottle yet. No need to introduce them now and completely break compatibility for older python versions.

SergBobrovsky:
I thought the minimum python version for bottle would be 3.6
f-string works from 3.6
ok, no more f-strings

PR #1286, opened 2021-01-01
compactize make_default_app_wrapper calls

SergBobrovsky:


bottle.py
```diff
@@ -3196,18 +3196,10 @@ def wrapper(*a, **ka):
...
-put       = make_default_app_wrapper('put')
-delete    = make_default_app_wrapper('delete')
-patch     = make_default_app_wrapper('patch')
-error     = make_default_app_wrapper('error')
-mount     = make_default_app_wrapper('mount')
-hook      = make_default_app_wrapper('hook')
-install   = make_default_app_wrapper('install')
-uninstall = make_default_app_wrapper('uninstall')
-url       = make_default_app_wrapper('get_url')
+(route, get, post, put, delete, patch, error, mount, hook,
```
RonRothman, line 3199: Your other cleanups are improvements, but this one in particular is a detriment to readability. Is there a benefit to changing this from something simple to something that requires a bunch of thinking?

It's more error prone, too: if someone simply doesn't realize that the order of names on lines 3199-3200 must be identical to the order of tokens in the string that spans lines 3201-3202, then they'll silently introduce an obscure bug.

RonRothman (changes requested):

bottle.py
oz123: I aggree with this comment. This current change maybe shorter, but it's error prone. In general I think a line with more than 5 assignments is a code smell. 
5 is the largest number we can visually process with out counting, so assignments that long are still comprehensible.

PR #1287, opened 2021-01-01
modified lines may be removed

SergBobrovsky:


bottle.py
```diff
@@ -4296,11 +4296,11 @@ def wrapper(*args, **kwargs):
 #: A dict to map HTTP status codes (e.g. 404) to phrases (e.g. 'Not Found')
 HTTP_CODES = httplib.responses.copy()
 HTTP_CODES[418] = "I'm a teapot"  # RFC 2324
-HTTP_CODES[428] = "Precondition Required"
-HTTP_CODES[429] = "Too Many Requests"
-HTTP_CODES[431] = "Request Header Fields Too Large"
-HTTP_CODES[451] = "Unavailable For Legal Reasons" # RFC 7725
-HTTP_CODES[511] = "Network Authentication Required"
+assert HTTP_CODES[428] == "Precondition Required"
```
RonRothman, line 4299: How about we just remove these assignments all together? I'm not sure that these `assert`s are really adding any value.

oz123: If I understand @SergBobrovsky, they are only there to convey that they can be removed.

The reason these can be removed is that [Python 3.5+ has a more complete status code collection](https://docs.python.org/3.5/library/http.html#http-status-codes) [compared to 2.7.](https://docs.python.org/2.7/library/httplib.html#httplib.HTTPS_PORT)

@SergBobrovsky  again, please document such changes in your git commit message. It's much easier to follow you then.
Also, as long as bottle supports Python 2.7 these lines should be left.  

Have they been added to `httplib`

PR #1292, opened 2021-01-02
Python 3 only: pyupgrade --py36-plus **/*.py

cclauss:
As discussed in #1193, this PR was created by running `pyupgrade --py36-plus **/*.py` to modernize Python syntax for Python 3 only usage.  https://github.com/asottile/pyupgrade  @defnull @SergBobrovsky

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 """
 Bottle is a fast and simple micro-framework for small web applications. It
 offers request dispatching (Routes) with URL parameter support, templates,
@@ -13,7 +12,6 @@ Copyright (c) 2009-2018, Marcel Hellkamp.
 License: MIT (see LICENSE for details)
 """
 
-from __future__ import print_function
 import sys
 
 __author__ = 'Marcel Hellkamp'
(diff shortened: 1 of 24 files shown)


defnull:
Locking out old python versions just because they are EOL is not the way to go. If source compatibility can be archived with low effort, why break it? Dropping support means that we won't test against that version any more and can remove annoying workarounds, but 3.5 is still used in a lot of places and I don't want to break source-compatibility 'just because'. That's also why I do not like automated tools in these situations. Some of these changes are unneeded and a human would solve these issue differently.

cclauss:
Asyncio is a bit more than “just because”.

oz123:
I think the approach of @defnull  here is reasonable. One of the best features of bottles is a small code base which changes slowly, which means its stable. The old syntax is fine, and as long as the maintainers of Python don't declare deprecated there is no reason to rush and refactor it.

defnull:
> Asyncio is a bit more than “just because”.

Am I missing something? What does asyncio have to do with this PR? I only see cosmetic changes that, unfortunately, break compatibility to older python versions without a clear benefit (other than cosmetics).

cclauss:
The link was just that continuing to support dead versions of Python means that newer features like asyncio are not available.   However, I am OK with the choice and with closing this PR.

PR #1294, opened 2021-01-02
refactor AppStack.default

SergBobrovsky:
To check `if self` is cheaper than make `try except` block

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2615,10 +2615,7 @@ class AppStack(list):
 
     @property
     def default(self):
-        try:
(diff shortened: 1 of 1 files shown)


bottle.py
```diff
@@ -2615,10 +2615,7 @@ def push(self, value=None):
 
     @property
     def default(self):
-        try:
-            return self[-1]
-        except IndexError:
-            return self.push()
+        return self[-1] if self else self.push()
```
oz123, line 2618: nice one!

PR #1300, opened 2021-01-03
remove redundunt check in static_file()

SergBobrovsky:


bottle.py
```diff
@@ -2883,7 +2883,7 @@ def static_file(filename, root,
 
     if not filename.startswith(root):
         return HTTPError(403, "Access denied.")
-    if not os.path.exists(filename) or not os.path.isfile(filename):
+    if not os.path.isfile(filename):
```
oz123, line 2886: Indeed, this is redundant.

```
In [71]: os.path.isfile("no-such-file")
Out[71]: False
```
@SergBobrovsky I believe that it will be easier for @defnull  to review all your PRs if your commit messages are a bit more verbose.
Adding stuff like this in the commit message surely helps.

RonRothman (approved):
LGTM

PR #1301, opened 2021-01-03
cosmetic improvement

SergBobrovsky:


Jwink3101:
I have no say in the matter but I think importing functions from a module goes against the "Zen of Python". Keep them in their own name space!

SergBobrovsky:
> I have no say in the matter but I think importing functions from a module goes against the "Zen of Python". Keep them in their own name space!

@Jwink3101 I believe that b64encode and b64decode are very specific words that cause name conflicts.
How would you rate this:
```
from unicodedata import normalize
```

RonRothman (approved):
LGTM

PR #1303, opened 2021-01-03
remove redundant `return None`

SergBobrovsky:


bottle.py
```diff
@@ -1296,13 +1296,11 @@ def json(self):
...
-            try:
-                return json_loads(b)
-            except (ValueError, TypeError):
-                raise HTTPError(400, 'Invalid JSON')
-        return None
+            if b:
+                try:
+                    return json_loads(b)
+                except (ValueError, TypeError):
+                    raise HTTPError(400, 'Invalid JSON')
```
oz123, line 1303: This is indeed shorter and clearer.

PR #1304, opened 2021-01-03
remove useless `lambda` wraper

SergBobrovsky:


RonRothman (approved):
LGTM

SergBobrovsky:
There is one problem: this "useless lambda" is a workaround to bypass the fact, that http_date() called on line 1807, and defined on 2962.
@defnull, what about  to lift `HTTP Utilities and MISC` section up to line ~200?

PR #1305, opened 2021-01-05
refactor Bottle._cast()

SergBobrovsky:
Standart method faster then lambda

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1090,10 +1090,10 @@ class Bottle(object):
         if isinstance(first, HTTPResponse):
             return self._cast(first)
         elif isinstance(first, bytes):
-            new_iter = itertools.chain([first], iout)
+            new_iter = itertools.chain((first,), iout)
         elif isinstance(first, unicode):
-            encoder = lambda x: x.encode(response.charset)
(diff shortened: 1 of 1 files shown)


bottle.py
```diff
@@ -1090,10 +1090,10 @@ def _cast(self, out, peek=None):
         if isinstance(first, HTTPResponse):
             return self._cast(first)
         elif isinstance(first, bytes):
-            new_iter = itertools.chain([first], iout)
+            new_iter = itertools.chain((first,), iout)
```
oz123, line 1093: IMHO this belongs in a separate commit with some more explanation.

bottle.py
```diff
@@ -1090,10 +1090,10 @@ def _cast(self, out, peek=None):
...
         elif isinstance(first, unicode):
-            encoder = lambda x: x.encode(response.charset)
-            new_iter = imap(encoder, itertools.chain([first], iout))
+            new_iter = map(str.encode, itertools.chain((first,), iout),
+                           itertools.repeat(response.charset))
```
oz123, line 1096: :100:

PR #1306, opened 2021-01-05
refactot Route.all_plugins()

SergBobrovsky:
My code accurately reproduces original functional, but is it correct?
In original code `unique` collects plugin names to avoid repeats. If plugin has no name, this check not. So, noname plugin can repeats. Is it right logic?

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -550,14 +550,14 @@ class Route(object):
 
     def all_plugins(self):
         """ Yield all Plugins affecting this route. """
-        unique = set()
-        for p in reversed(self.app.plugins + self.plugins):
-            if True in self.skiplist: break
-            name = getattr(p, 'name', False)
-            if name and (name in self.skiplist or name in unique): continue
-            if p in self.skiplist or type(p) in self.skiplist: continue
-            if name: unique.add(name)
-            yield p
+        if True in self.skiplist: return
+        skips = set(self.skiplist)
+        for plugins in self.plugins, self.app.plugins:
+            for p in reversed(plugins):
+                if p not in skips and type(p) not in skips:
+                    name = getattr(p, 'name', False)
+                    if name and name not in skips: skips.add(name)
+                    yield p
 
     def _make_callback(self):
         callback = self.callback
--- a/bottle.py
+++ b/bottle.py
@@ -552,11 +552,11 @@ class Route(object):
         """ Yield all Plugins affecting this route. """
         if True in self.skiplist: return
         skips = set(self.skiplist)
-        for plugins in self.plugins, self.app.plugins:
-            for p in reversed(plugins):
-                if p not in skips and type(p) not in skips:
-                    name = getattr(p, 'name', False)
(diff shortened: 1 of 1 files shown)


bottle.py
```diff
@@ -550,14 +550,14 @@ def prepare(self):
...
-            if name and (name in self.skiplist or name in unique): continue
-            if p in self.skiplist or type(p) in self.skiplist: continue
-            if name: unique.add(name)
-            yield p
+        if True in self.skiplist: return
+        skips = set(self.skiplist)
+        for plugins in self.plugins, self.app.plugins:
+            for p in reversed(plugins):
+                if p not in skips and type(p) not in skips:
+                    name = getattr(p, 'name', False)
```
oz123, line 558: Do you really need `False` here ? `getattr` will return `None` by default.

bottle.py
```diff
@@ -550,14 +550,14 @@ def prepare(self):
...
-            if True in self.skiplist: break
-            name = getattr(p, 'name', False)
-            if name and (name in self.skiplist or name in unique): continue
-            if p in self.skiplist or type(p) in self.skiplist: continue
-            if name: unique.add(name)
-            yield p
+        if True in self.skiplist: return
+        skips = set(self.skiplist)
+        for plugins in self.plugins, self.app.plugins:
+            for p in reversed(plugins):
```
oz123, line 556: Compared to the original code I find this much harder to understand.
normally a `for` loop in python is in the form:
```
for single in plural
    do something ... 
```
You wrote for `plugins` in `<other list of plugins>`. I didn't dive into analyzing in depth to see what is the other set of plugins. But this form of `for` will immediately cause a headache for a lot of people.
Further, an embedded for loop is harder to understand compared to a simple for loop.

bottle.py
Marrin: @oz123 `getattr()` raises `AttibuteError` by default!  So there needs to be a third argument, but IMHO it should be `None` and not `False`.

oz123:
@SergBobrovsky you do a really cool job on those PRs. Keep up and take the critics as constructive criticism.

bottle.py
SergBobrovsky: @oz123, you wrong, if len(args) == 2 and obj has no attr, error rises:
```
getattr(None, 'name')
  File "/usr/lib/python3/dist-packages/IPython/core/interactiveshell.py", line 3331, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-41470b2f07a7>", line 1, in <module>
    getattr(None, 'name')
AttributeError: 'NoneType' object has no attribute 'name'
```

oz123: Yeah, my bad one. Mixed it `dict.get`!

oz123 (commented):
Much better now.

Neat: Fix the typo in the PR's title:

 refactot Route.all_plugins() #1306  ->  refactor Route.all_plugins() #1306

PR #1312, opened 2021-01-08
refactor Route.get_undecorated_callback()

SergBobrovsky:
`while` loop improved, made more readable.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -577,19 +577,15 @@ class Route(object):
         """ Return the callback. If the callback is a decorated function, try to
             recover the original function. """
         func = self.callback
-        func = getattr(func, '__func__' if py3k else 'im_func', func)
         closure_attr = '__closure__' if py3k else 'func_closure'
         attributes = getattr(func, closure_attr, False)
         while attributes:
-            func = attributes[0].cell_contents
-            attributes = getattr(func, closure_attr, False)
-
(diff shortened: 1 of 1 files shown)


bottle.py
```diff
@@ -579,9 +579,10 @@ def get_undecorated_callback(self):
         func = self.callback
         func = getattr(func, '__func__' if py3k else 'im_func', func)
         closure_attr = '__closure__' if py3k else 'func_closure'
-        while hasattr(func, closure_attr) and getattr(func, closure_attr):
-            attributes = getattr(func, closure_attr)
+        attributes = getattr(func, closure_attr, False)
+        while attributes:
```
oz123, line 583: The original code does not define `attributes` if ` hasattr(func, closure_attr)` isn't true.
Your change always defines it.
I suspect (have not benchmarked it) that ` hasattr(func, closure_attr)` is faster than the creation of `attribute = Flase`. Hence, I would rather keep the original code untouched.

SergBobrovsky: get_undecorated_callback() serves to one goal - to be called for repr(callback), ie for debug via `print()`. Nothing critical for speed.

SergBobrovsky:
What means `__func__` attribute in original `get_undecorated_callback`? May be someone thought about `__call__` attribute, but wrote `__func__`? If so, some change requared.

PR #1315, opened 2021-01-08
refactor BaseRequest.remote_route()

SergBobrovsky:
map with native function faster than list comprehension

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1527,7 +1527,7 @@ class BaseRequest(object):
             work if all proxies support the ```X-Forwarded-For`` header. Note
             that this information can be forged by malicious clients. """
         proxy = self.environ.get('HTTP_X_FORWARDED_FOR')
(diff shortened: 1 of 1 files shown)


bottle.py
```diff
@@ -1527,7 +1527,7 @@ def remote_route(self):
             work if all proxies support the ```X-Forwarded-For`` header. Note
             that this information can be forged by malicious clients. """
         proxy = self.environ.get('HTTP_X_FORWARDED_FOR')
-        if proxy: return [ip.strip() for ip in proxy.split(',')]
+        if proxy: return list(map(str.strip, proxy.split(',')))
```
RonRothman, line 1530: IMO, the list comprehension is (a) more Pythonic, and (b) easier on the eyes. I recommend keeping it the prior way.

oz123: I agree here.

PR #1316, opened 2021-01-09
refactor ServerAdapter.__repr__()

SergBobrovsky:
"%r" placeholder exists at least from 2008
https://www.python.org/dev/peps/pep-3140/


oz123 (commented):
Nice one. I didn't know that `%r` can be used instead of `%s`.

Issue #1115, opened 2018-12-13
path is None


shakna-israel:
This seems to be related to some of the work coming out of #1094.

```bottle==0.12.13```

```python==3.7.1```

Here's my little startup script, (```app``` is a WSGI application. The ```default_app``` with a couple wrappers).

```
import os

try:
    port = os.environ['PORT']
    server='twisted'
except KeyError:
    port = 5000
    server=False

if not port:
    port = 5000

if server:
    bottle.run(server=server, port=port, app=app, host='0.0.0.0', reloader=True)
else:
    bottle.run(port=port, app=app, host='0.0.0.0', reloader=True)
```

This, in turn does run, but the ```reloader``` part seems to cause some problems, though I'm not quite sure how, as the server continues to run and perform as expected.

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/home/jmilne/mindfulkitchen/venv/lib/python3.7/site-packages/bottle.py", line 3158, in run
    if path[-4:] in ('.pyo', '.pyc'): path = path[:-1]
TypeError: 'NoneType' object is not subscriptable
```

So at that point in ```run```, ```path``` is ```None``` for some reason.

I know I am using a newer Python than is tested, and the app is actually quite sizeable, so I might have caused something to go wrong somewhere higher up in the stack, or it might be some incompatibilities with ```Twisted```.

defnull:
The reloader scans all loaded modules to detect changes. Each module
should have its source file stored in `module.__file__`. On of your
dependencies has this attribute set to `None`, which should not be
allowed according to the python spec:

> |__file__| is optional. If set, this attribute’s value must be a
string. The import system may opt to leave |__file__| unset if it has no
semantic meaning (e.g. a module loaded from a database).

So, __file__ being set, and not a string, is a bug in the dependency I
think. Can you find out which dependency causes the error?

for name, module in sys.modules.items():
    if getattr(module, '__file__', '') is None:
        print("Module with broken __file__: ", name)

We can work around this in bottle, but it would be even better to fix it
in the actual module that causes this error.


Am 13.12.18 um 06:26 schrieb James Milne:

shakna-israel:
That would be ```google-api-python-client``` and one of it's dependencies, ```zope.interface```.

I guess I'll chase it up with Google then.

shakna-israel:
[...]

PR #1317, opened 2021-01-09
remove some redundant from FileCheckerThread.run()

SergBobrovsky:
third arg of `getattr` is default value


oz123 (commented):
Good catch.

defnull:
Turns out that `or ''` is needed if `__file__` is `None`, which should never happen but does happen. See #1455 and #1115

PR #1333, opened 2021-02-13
Redundant system separation in search method

felipexlr50:
The os.path.abspath() already output with the separation in the
end of the string, therefore the os.sep() is redundant and can output
missing file when search for template file.


defnull:
No it does not?
```>>> os.path.abspath("/foo/bar/")
'/foo/bar'```

PR #1346, opened 2021-07-14
Fix wsgiserver import error for cherrypy server

ibcode15:
wsgiserver now has to be imported from cheroot.wsgi not cherrypy

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -3292,7 +3292,10 @@ class CherryPyServer(ServerAdapter):
         depr(0, 13, "The wsgi server part of cherrypy was split into a new "
                     "project called 'cheroot'.", "Use the 'cheroot' server "
                     "adapter instead of cherrypy.")
-        from cherrypy import wsgiserver # This will fail for CherryPy >= 9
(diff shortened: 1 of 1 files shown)


bottle.py
```diff
@@ -3292,7 +3292,10 @@ def run(self, handler):  # pragma: no cover
         depr(0, 13, "The wsgi server part of cherrypy was split into a new "
                     "project called 'cheroot'.", "Use the 'cheroot' server "
                     "adapter instead of cherrypy.")
-        from cherrypy import wsgiserver # This will fail for CherryPy >= 9
+        try:
+            from cherrypy import wsgiserver
+        except ImportError:
+           from cheroot.wsgi import Server as wsgiserver
```
jneeee, line 3298: Indent error

jneeee (changes requested):

defnull:
There is a separate `CherootServer` adapter just below the `CherryPyServer` class, but we can think about removing `CherryPyServer` in 0.14

PR #1354, opened 2021-10-11
Fix handling of empty If-Modified-Since header for static files

braedon:
Previously, if the If-Modified-Since header was an empty string it wouldn't be parsed into an int (or None), but would still be compared to the int mtime, causing a TypeError.


defnull:
Not sure if an empty `If-Modified-Since` is valid in the first place, but broken clients are common enough to address this. Thanks!

PR #1355, opened 2021-10-17
Removed a pitfall where the code was using a default mutable argument

NaelsonDouglas:
**The problem**
In Python it usually dangerous to use mutable arguments like dicts or lists as default arguments in method, as is better explained [here](https://docs.python-guide.org/writing/gotchas/)

**the solution**
This PR applies a simple refactoring to remove a case where a method was created using default a mutable argument


defnull:
The `globals` variable is not mutated and does not leave the local scope of the function. This is safe to do.
The other change does not not seem to have any effect.

PR #1360, opened 2021-12-03
Fix #1359: check for attribute using hasattr

### LINKED ISSUE #1359: Setting custom request attributes requires name validation
https://github.com/bottlepy/bottle/blob/71bb933649ba009cd537c2e973519d6e024c02a2/bottle.py#L1594-L1599

What about `request.body = 'blah'`?
By the idea it should raise an error, since `body`  is read-only property, but it doesn't, it just creates `environ['bottle.request.ext.body']`.
Note that `request.body` will still return 'real body' (BytesIO), not `'blah'`

This issue affects all request attributes/properties.

nootr:
This PR fixes a bug where setting an existing, read-only property of a BaseRequest object does not raise an AttributeError.

Fixes #1359

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1594,7 +1594,7 @@ class BaseRequest(object):
     def __setattr__(self, name, value):
         if name == 'environ': return object.__setattr__(self, name, value)
         key = 'bottle.request.ext.%s' % name
-        if key in self.environ:
+        if hasattr(self, name):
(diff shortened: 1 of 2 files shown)


avelino (approved):
LGTM

defnull:
Nice catch, thanks!

PR #1363, opened 2022-01-05
fix #1361 inclusive language comment

### LINKED ISSUE #1361: Update comment for inclusive language
Hi there,

I would like to suggest the updating of bottle.py for inclusive language. It's one small word in one comment here:

https://github.com/bottlepy/bottle/blob/71bb933649ba009cd537c2e973519d6e024c02a2/bottle.py#L1658

At Artsy, we're updating our repositories and dependencies here per this RFC artsy/README#427. For such a change we are adopting 'denylist' over 'blacklist.'

For such a small change I would normally just open a PR, but I want to respect community norms and am new here. 
Thanks. KT

kathytafel:
Fixed: #1361
See issue for context.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1655,7 +1655,7 @@ class BaseResponse(object):
     default_status = 200
     default_content_type = 'text/html; charset=UTF-8'
 
-    # Header blacklist for specific response codes
+    # Header denylist for specific response codes
     # (rfc2616 section 10.2.3 and 10.3.5)
     bad_headers = {
         204: frozenset(('Content-Type', 'Content-Length')),


avelino (changes requested):
I need you to review the test suite, if you changed "only" comment, but the tests broke

defnull:
Nah, thee tests break all the time. These server-adapter tests are very unstable unfortunately. I kinda regret adding this feature instead of just documenting how to use the different servers.

kathytafel:
Thanks all!!

avelino:
@kathytafel thank you for bringing a more inclusive view to the bottle

PR #1370, opened 2022-04-14
Add an optional redirect(cache=sec) argument

da2x:
The argument sets an optional `Cache-Control` header for redirects. Allows for more fined-grained control over client and browser caching for known short- and long-lived redirects. The argument accepts an integer value with the number of seconds to add to the `max-age` directive.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2813,13 +2813,15 @@ def abort(code=500, text='Unknown Error.'):
     raise HTTPError(code, text)
 
 
-def redirect(url, code=None):
+def redirect(url, code=None, cache=None):
     """ Aborts execution and causes a 303 or 302 redirect, depending on
         the HTTP protocol version. """
     if not code:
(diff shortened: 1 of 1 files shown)


bottle.py
```diff
@@ -2813,13 +2813,15 @@ def abort(code=500, text='Unknown Error.'):
...
 
-def redirect(url, code=None):
+def redirect(url, code=None, cache=None):
     """ Aborts execution and causes a 303 or 302 redirect, depending on
         the HTTP protocol version. """
     if not code:
         code = 303 if request.get('SERVER_PROTOCOL') == "HTTP/1.1" else 302
     res = response.copy(cls=HTTPResponse)
     res.status = code
+    if type(cache) is int:
```
avelino, line 2823: You can check if the variable exists, you don't need to look at its type.
If it is an unexpected string it will be the developer's responsibility

avelino (changes requested):
We need to write test validating the functioning of your implementation

PR #1382, opened 2022-06-13
Was getting unsafe io error.

Farrenzo:
changed to safe file view. Please provide feedback.


defnull:
That change would read the entire files content into memory, which is not an option.

Actually the WSGI server should close the response object if it has a `close` method, which is ensured by bottle for file based responses. You should not get that warning (with bottle.py from master branch, not sure about 0.12). Can you build a reproducer and open an Issue please?

defnull:
I found (and fixed) a situation (requests with `Accept-Range` header) where the close() method is masked. Perhaps this already gets rid of the warning for you.

PR #1390, opened 2022-08-14
Fix the link for fapws project.

girisagar46:
My oh my. Thank goodness I clicked this link while I was at home. It would be disastrous if I were in the office or public space.

⚠️ NSFW Warning: Please **DO NOT** go to http://www[dot]fapws[dot]org/ if you're in a public space.


defnull:
What the ... thanks.

Issue #1393, opened 2022-09-05
Python 3.8+ missing from Pypi Programming Language


clavedeluna:
[pypi](https://pypi.org/project/bottle) bottom left in Programming Language section is missing Python 3.8, 3.9 (3.10, too?) support. I can see they are in [setup.py](https://github.com/bottlepy/bottle/blob/master/setup.py#L41) so the issue is likely a release issue.

Side question: are releases done manually?

clavedeluna:
Great! Added Py 3.10 to master. Once that's merged and master is released, Pypi should be up to date.

PR #1395, opened 2022-09-05
add Python 3.10 to pypi section of setup.py

clavedeluna:
Handles #1393 

add Python 3.10 to pypi list.


defnull:
This should be solved by now. Thanks :)

PR #1396, opened 2022-09-05
rename TestError to ExampleError to avoid PytestCollectionWarning

clavedeluna:
When running tests, this warning is raised:

```
test/test_exc.py:4
  /Users/dani/bottle/test/test_exc.py:4: PytestCollectionWarning: cannot collect test class 'TestError' because it has a __init__ constructor (from: test/test_exc.py)
    class TestError(Exception):
```

this is because pytest is trying to collect all classes starting with the word `Test`. Since this class isn't an actual test, renaming it to something else is appropriate. 
Another solution would've been to add `  __test__ = False` to the class, but renaming seemed straight-forward.

It's always nice to run tests and not see warnings, so I may fix some more in upcoming PRs :)

### DIFF
--- a/test/test_exc.py
+++ b/test/test_exc.py
@@ -1,7 +1,7 @@
 import bottle
 from .tools import ServerTestBase
 
-class TestError(Exception):
+class ExampleError(Exception):
     pass
 
 class TestAppException(ServerTestBase):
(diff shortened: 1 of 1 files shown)


defnull:
This should be solved by now. Thanks :)

Issue #1195, opened 2020-01-03
Add flake8 (with dlint) tests to CI


defnull:
#1194 showed that adopting properly configured static analysis might help prevent some serious bugs. There are lots of style warnings and false-positives currently when running flake8. Some are fixable, but some warnings could also simply be disabled.

mschwager:
Hey, glad to see this change!

Regarding false-positives from `flake8`: if you simply want to include Dlint to search for potential security issues and bugs you can run `python -m flake8 --select=DUO`, to run all `flake8` checks, including stylistic ones, you can omit the `--select=DUO`.

Generally I find running all the `flake8` checks useful for conforming to a certain style (i.e. PEP8), and finding some basic issues like unused imports and unused variables. If you'd like to quickly move your codebase to `flake8` conformance I'd take a look at [autopep8](https://pypi.org/project/autopep8/) and [autoflake](https://pypi.org/project/autoflake/).

Regarding Dlint's other findings:

```
$ python -m flake8 --select=DUO bottle.py 
bottle.py:160:5: DUO105 use of "exec" is insecure
bottle.py:160:10: DUO110 use of "compile" is insecure
bottle.py:2922:16: DUO130 insecure use of "hashlib" module
bottle.py:3616:12: DUO104 use of "eval" is insecure
bottle.py:3965:16: DUO110 use of "compile" is insecure
bottle.py:4007:9: DUO105 use of "exec" is insecure
```

If you don't believe these are security issues, then I'd [ignore them inline](https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors).

Happy to help here if you'd like!

PR #1399, opened 2022-09-06
add flake8 linting to Makefile and github actions

clavedeluna:
Related to #1195

I saw #1195 and since I have experience with flake8 I decided to contribute.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+max-line-length = 120
+exclude =
+	.tox
+	.git
--- a/.github/workflows/install-deps.sh
+++ b/.github/workflows/install-deps.sh
@@ -3,7 +3,7 @@ set -e
 set -x
 
 # Test utilities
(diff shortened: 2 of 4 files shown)


defnull:
I'm not a big fan of automates code formatting or, in this case, style checkers to be honest. Bottle is written by hand and most of the lines that trigger flake8 warnings (even after ignoring the 10 most common error-types) were written that way on purpose. I often 'break the rules' to transport information via formatting. For example: Code that should be uninteresting for the average reader (e.g. trivial implementation details) is written more compact so it does not distract from the actually important parts. Instead of having two screens of import statements, I put all of the obvious ones in one line. If a number of trivial methods of a class all fit in one line each, I write one-line def statements without lines in-between, using only one third of the space. And so on.

Adding a flake8 check for PRs would be dishonest, because I myself would ignore those warnings most of the time. I'd rather have PRs that fit the style of the surrounding code.

PR #1411, opened 2023-02-27
Adjusts the test_server 'fetch' method to disable proxying within test cases where it is used

### LINKED ISSUE #1410: 0.12.24: test_server: 'fetch' method may unintentionally use system proxy
The `test_server` test module performs a bunch of webserver self-tests against `localhost` (127.0.0.1), and includes a [`fetch` method definition](https://github.com/bottlepy/bottle/blob/e2404c311fca55ae5058b714db2349aa8eb1e6bd/test/test_server.py#L79-L83) to support that.

The (probably sensible) [defaults for `urlopen`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen) will use a proxy if one is configured in the environment:

> In addition, if proxy settings are detected (for example, when a `*_proxy` environment variable like `http_proxy` is set), [`ProxyHandler`](https://docs.python.org/3/library/urllib.request.html#urllib.request.ProxyHandler) is default installed and makes sure the requests are handled through the proxy.

...but since we're making a request to `localhost` as part of a unit test, we probably don't want the potential for remote network services to be involved (since they could affect test performance and results).  Fortunately we can disable them by using [`ProxyHandler.__init__` arguments](https://docs.python.org/3/library/urllib.request.html#urllib.request.ProxyHandler):

> To disable autodetected proxy pass an empty dictionary.

(and even better, for backwards-compatibility in this library: the [same parameter is available in Python 2.7's `urllib2.ProxyHandler`](https://docs.python.org/2.7/library/urllib2.html#urllib2.ProxyHandler))

jayaddison:
May resolve #1410 (under discussion).

### DIFF
--- a/test/test_server.py
+++ b/test/test_server.py
@@ -11,9 +11,9 @@ from . import tools
 from bottle import _e
 
 try:
-    from urllib.request import urlopen
+    from urllib.request import ProxyHandler, build_opener
 except:
-    from urllib2 import urlopen
+    from urllib2 import ProxyHandler, build_opener
 
 serverscript = os.path.join(os.path.dirname(__file__), 'servertest.py')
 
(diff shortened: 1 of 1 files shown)


defnull:
Merged (manually) and released. Thanks :)

jayaddison:
You're welcome - thank you!

PR #1414, opened 2023-03-09
auth_basic: decide in check function how to deal with missing creds

karmab:


defnull:
That's a breaking change in a sensitive area unfortunately. An app may have a check function that always returns True, for example because there is a middleware or proxy that handles SSO logins and adds or removes an auth header with an empty password depending on the users login state. That is actually quite common. Those apps would break and allow guests to access protected endpoints.

The `auth` decorator is really small, if you need to change it, just copy&paste the code and use your own version.

PR #1426, opened 2023-08-01
Avoid KeyError: 'PATH_INFO' errors

bulletmark:
I use bottle often and in many of the applications I get frequent tracebacks:
```
  File "/src/pvoproxy/venv/lib/python3.11/site-packages/bottle.py", line 863, in _handle
     path = environ['bottle.raw_path'] = environ['PATH_INFO']
                                         ~~~~~~~^^^^^^^^^^^^^
KeyError: 'PATH_INFO'
```

This PR fixes that.


defnull:
`PATH_INFO` is a required field in WSGI environments. It can be empty, but if it's actually missing, then that's a bug in the WSGI server you are using.

PR #1427, opened 2023-08-02
generated stub to avoid pyright errors

ilrico:
To avoid pyright errors
1. generated bottle.pyi stub (with stubgen)
2. then modified:

Bottle.route(method: str | list[str])
def json(self) -> dict | None: ...


defnull:
Correct me if I'm wrong, but last time I checked stub files for single-file modules were broken, because the required `py.typed` marker file cannot be placed into any package directory. This is an oversight in the standard PEPs, see https://github.com/python/typing/issues/1333. Setuptools also refuses to put `bottle.pyi` into the binary wheel distribution. I tried, but could not find a way to properly package bottle with the stub file.

That said, development on `0.14-dev` started and we can drop support for Python < 3.8 now, so instead of a stub file we could (slowly) start annotating directly in bottle.py. That might be the better solution in the long run.

PR #1436, opened 2023-11-28
using add exec_module in the _ImportRedirect

### LINKED ISSUE #1435: `_ImportRedirect` is using deprecated `load_module` function
It's been deprecated in https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module , and runnint the unittest via ` python -m unittest -k "TestImportHook"` will got this warning:

```
<frozen importlib._bootstrap>:671: ImportWarning: _ImportRedirect.exec_module() not found; falling back to load_module()
```

aisk:
fix #1435

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2050,8 +2050,10 @@ class _ImportRedirect(object):
             '__loader__': self
         })
         sys.meta_path.append(self)
+        self.fullname = None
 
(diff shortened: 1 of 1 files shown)


defnull:
Thanks for the PR, I managed to fix it without the need for temporary state in the loader. The full name is part of the ModuleSpec.

Issue #1403, opened 2022-11-07
"cgi" standard library module used by bottle is deprecated in Python 3.11, to be removed in 3.13


liyanage:
I'm running bottle with Python 3.11 for the first time and I see this deprecation warning:

```
/xxx/lib/bottle/bottle.py:72: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
  import base64, calendar, cgi, email.utils, functools, hmac, itertools,\
```

I guess bottle needs to stop using this module?

defnull:
That's unfortunate, bottle uses the `form-data/multipart` parser from that library. The PEP says:

> FieldStorage/MiniFieldStorage has no direct replacement, but can typically be replaced by using [multipart](https://pypi.org/project/multipart/) (for POST and PUT requests) or urllib.parse.parse_qsl (for GET and HEAD requests)

Great. I'm the author of `multipart`, so I'm okay with that, but copy&pasting all that into bottle (to maintain the zero-dependency design goal) feels wrong.

zb3:
Removing `multipart/form-data` parser from Python standard library is unthinkable. I understand the `cgi` module was deprecated because CGI is itself deprecated, but the `multipart` encoding is not deprecated, it's still used today, hence I totally don't understand this..

simsong:
[PEP 594](https://peps.python.org/pep-0594/): Remove the cgi and cgitb modules, deprecated in Python 3.11.

cgi.FieldStorage can typically be replaced with [urllib.parse.parse_qsl()](https://docs.python.org/3.13/library/urllib.parse.html#urllib.parse.parse_qsl) for GET and HEAD requests, and the [email.message](https://docs.python.org/3.13/library/email.message.html#module-email.message) module or [multipart](https://pypi.org/project/multipart/) PyPI project for POST and PUT.

(per https://docs.python.org/3.13/whatsnew/3.13.html)

fedepell:
Any news maybe on this? 3.13 is now getting closer with 3.12 being out in the meantime, albeit the release plan is around half of next year.

defnull:
The only viable solution is to copy&paste code from `multipart` into bottle. I really really do not want to do that, but I have no other idea how to maintain the single-file no-deps aspect any other way.

fedepell:
> The only viable solution is to copy&paste code from `multipart` into bottle. I really really do not want to do that, but I have no other idea how to maintain the single-file no-deps aspect any other way.

[...]

PR #1437, opened 2023-11-28
Using email module to parse multipart insteal of the deprecated cgi module

aisk:
fix: #1403

Since `cgi` will be removed, the Python change log recommends to using `email.message` or the PYPI package `multipart`, and bottle does not allow to use external dependencies, and vendoring `multipart` is not a good practice, so I think the `email` package is a better way.

I don't check too much about the compatibilities, if some maintainer think this way is okay, I'll invest more time to do it. But the `test_multipart` passed on my local machine (some other tests failed because I'm using Windows and they failed in the master branch).

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -69,7 +69,7 @@ if __name__ == '__main__':
 # Imports and Python 2/3 unification ##########################################
 ###############################################################################
 
-import base64, calendar, cgi, email.utils, functools, hmac, itertools,\
+import base64, calendar, email.utils, functools, hmac, itertools,\
        mimetypes, os, re, tempfile, threading, time, warnings, weakref, hashlib
 
 from types import FunctionType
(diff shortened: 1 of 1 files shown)


defnull:
Unfortunately, all data parsed by `email.parser.FeedParser` will end up in memory buffered `Message`s. Uploading large files (or many small ones) would likely trigger `MemoryError` on a busy server. The parser needs a way to offload large file uploads into temporary files to be useful in a web context. Not sure if the `email.parser` package supports that use case.

aisk:
The `email.parser.FeedParser` has an optional argument `_factory`, which can specify which `Message` class will be used in the parsed result.  So we can subclass the `email.message.Message`, and override the `set_payload` and any other methods to offload the large file to the disk.

I didn't take too much to see if this will work, if you didn't check this too, I want  to investigate on it.

defnull:
Does not really help for large uploads, as those are still collected as a list of strings in memory before `set_payload` is even called. That behavior is hard-coded in the parser. The parser is also string-based, binary data is passed in as `data.decode('ascii', 'surrogateescape')` and copied multiple times. It was designed for emails (where you need an error tolerant and lax parser) and not for the internet (where you need a fast and strict parser that bails immediately if it sees something fishy). I would love to use that parser, that would be my first choice if that was an option. But I do not think it is suitable for this use case.

aisk:
Thanks for the kindly reply, I've got the point!

defnull:
0.13 was released with a full-featured solution. Thanks :)

PR #1440, opened 2024-02-02
Update GitHub workflow with more recent versions of Python and GitHub actions.

vsajip:


defnull:
release-0.13 and master should have correct versions now. Thanks :)

PR #1453, opened 2024-09-09
handle case when WSGIHeaderDict has None in values

### LINKED ISSUE #1452: WSGIHeaderDict fails to get item if it's present and None in original dict
get key with not None value
```
>>> import bottle
>>> x = bottle.WSGIHeaderDict({"HTTP_X": "y", "HTTP_Z": None})
>>> print(x.get("x"))
y
```

get absent key
```
>>> print(x.get("w"))
None
```

get key with None value
```
>>> print(x.get("z"))
  File "/usr/lib64/python3.6/_collections_abc.py", line 660, in get
    return self[key]
  File "/root/work/build/debug/config/vnc_openstack/.tox/py36/lib/python3.6/site-packages/bottle.py", line 2331, in __getitem__
    val = val.decode('utf8')
AttributeError: 'NoneType' object has no attribute 'decode'
```

Andrey-mp:
if key has None value then .get(key) throws an Exception

fixes #1452

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -2307,7 +2307,7 @@ class WSGIHeaderDict(DictMixin):
 
     def __getitem__(self, key):
         val = self.environ[self._ekey(key)]
-        if py3k:
+        if val and py3k:
             if isinstance(val, unicode):
                 val = val.encode('latin1').decode('utf8')
             else:


defnull:
As discussed in the issue, `None` is not a valid value type for a WSGI header, it should never appear in `environ` and never be returned as a header value from Bottle. Crashing is the correct behavior if the `environ` does not meet expectations.

PR #1454, opened 2024-09-10
Include the captured exception on invalid JSON

sirosen:
I encountered this as an issue while handling the bottle 0.13 update from 0.12.
The behavior of JSON decoding failures has changed slightly vs 0.12, and while updating I was surprised to find that `error.exception` was unset, when a JSONDecodeError was caught.

Setting the inner exception to the caught exception helps callers catch and re-emit JSON decoding errors with their own code (e.g., to meet error API standards for their application).

---

Please let me know if this PR is missing anything it needs to progress.
I'm happy to take extra steps (add new tests, sign a CLA, etc) as appropriate.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1277,8 +1277,8 @@ class BaseRequest(object):
                 return None
             try:
                 return json_loads(b)
-            except (ValueError, TypeError):
-                raise HTTPError(400, 'Invalid JSON')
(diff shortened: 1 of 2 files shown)


defnull:
Thanks, that makes a lot of sense. No CLA needed (Bottle is MIT Licensed). I'll check this tomorrow and merge :)

PR #1473, opened 2025-02-16
Update tutorial_app.rst

noisefloor:
Complete re-write of the tutorial app. App itself state the same, improved the Python code of the app, updated explanations, complete updated of deployment section of the tutorial.


defnull:
Wow, thanks for your work! This is great!

PR #1479, opened 2025-03-14
Improve static_file

guangrei:
Implemented `_rangeiter` not only for the `HTTP_RANGE` context. This improves serving large files and memory efficiency.


defnull:
Actual file objects can be heavily optimized by the WSGI server and should not be wrapped in an iterator. See https://github.com/bottlepy/bottle/blob/master/bottle.py#L1007

PR #1482, opened 2025-03-22
Make FileUpload.save accept Path instance for destination

johndoe46:
Path instance is detected and converted to str. No other changes.


defnull:
I would prefer to actually move from `os.path` to `pathlib.Path` whenever possible, ans support both `str` and `os.PathLike` as input to those functions/methods.

PR #1485, opened 2025-04-13
Fixed ResourceWarning unclosed file?

nayuki:
I'm running a Bottle-based web server application, which contains a route like this:

```
@bottle.get("/media.mp4")
def _get():
    return bottle.static_file("media.mp4", ".")
```

Randomly about a few times a day, a rare warning messages gets printed to the console that is running Python with Bottle and the application, something like this:

```
/home/user/website/bottle.py:1860: ResourceWarning: unclosed file <_io.BufferedReader name='/home/user/website/video.mp4'>
  ls.var = value
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

After tracing the flow of objects through a lot of code within Bottle, flup (FastCGI), and the Python standard library's WSGI, as well as implementing my own WSGI server, and also doing some stress testing on the server where I hit hard-reload on my browser many times in quick succession, I think I narrowed down the problem.

In this Bottle code, if an exception is raised during the call to `start_request()` (e.g. the underlying socket to write to has been abruptly closed by the other end), and the object returned from the application was a resource (which needs to be closed) and not simply some bytes in memory (which can be garbage-collected without drama), then Bottle leaks the resource. Relevant pieces of code:

https://github.com/bottlepy/bottle/blob/2fa7afd35feb8f4c67699b670cc45556e93f62c2/bottle.py#L1044-L1072

https://github.com/bottlepy/bottle/blob/2fa7afd35feb8f4c67699b670cc45556e93f62c2/bottle.py#L2719-L2826

My one-line patch seems to make things better when testing in production and in stress-testing, but because I'm unfamiliar with the entirety of Bottle's architecture, I am not 100% sure that this is a complete fix or the most appropriate change. Thank you for your consideration.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1058,6 +1058,7 @@ class Bottle(object):
         except (KeyboardInterrupt, SystemExit, MemoryError):
             raise
         except Exception as E:
+            if hasattr(out, 'close'): out.close()
             if not self.catchall: raise
             err = '<h1>Critical error while processing request: %s</h1>' \
                   % html_escape(environ.get('PATH_INFO', '/'))


defnull:
Tests fail because `out` may not be defined at that point if `self._cast(self._handle(environ))` raises an exception, but you may still have found and fixed an issue that I was never able to reproduce when it was reported. Great find!

But I'm still a bit confused why this can happen. Calling `start_response` should not throw an exception, even if the client closed the socket, unless your response is actually empty (`Content-Length: 0`). But empty responses usually do not need to be closed. PEP-3333 says:

> However, the start_response callable must not actually transmit the response headers. Instead, it must store them for the server or gateway to transmit only after the first iteration of the application return value that yields a non-empty bytestring, or upon the application’s first invocation of the write() callable. In other words, response headers must not be sent until there is actual body data available, or until the application’s returned iterable is exhausted. (The only possible exception to this rule is if the response headers explicitly include a Content-Length of zero.)

Which means one three three things:
* The WSGI server is buggy and tries to write headers to the socket immediately.
* The WSGI server checks the state of the socket within `start_response` and throws, which is not forbidden by PEP-3333 but also not very common.
* You return a close-able response that also has a `Content-length: 0` header or answers to a `HEAD` request. `HEAD` requests are handled by `static_file`, but empty files are not (yet).

nayuki:
Thanks for the really helpful explanations, and sorry I didn't do enough work on my part.

You are right; `start_response()` is only supposed to store the headers and not perform any I/O, so it should not be raising an exception. I actually read [PEP-3333](https://peps.python.org/pep-3333/) in full before tackling this issue, but kind of forgot some key details along the way. Your reminder about the contract semantics was very relevant.

The fact that the automated tests fail surprised me, because my code change seems to work well in production. But yes, it is technically not kosher because the evaluations of `self._handle(...)` and `self._cast(...)` can raise an exception before the local variable `out` is assigned.

There is a sub-issue here that I glossed over initially. My complaint about `ResourceWarning` shows up when using the flup WSGI server, which presumably has stood the test of time. But it occurs rarely, and I'm not sure how to trigger it reliably in a deliberate way. Mashing reload and also having some streaming media files (e.g. big videos) half-open seem to help. This is something I've noticed for a decade, but never had the time to do anything about it until a few months ago when I dug deeply into flup to see if I could improve the library.

The `ResourceWarning` got much worse on my own brand new FastCGI-WSGI server, where I did not implement `start_response()` in the compliant deferred-I/O manner you described. The patch does help this case and it is easy to trigger, but it turns out it doesn't seem to help in general. I re-tested the flup server even with this one-line Bottle patch, and it didn't seem to help. Sorry, but it looks like I'm no closer to figuring out where exactly the file objects are getting leaked.

If you have any general comments or pointers to where I should investigate next, I would appreciate it. Otherwise, I'm fine with closing the issue as a false alarm.

defnull:
I spent some time and found a couple of edge cases that can cause the close method to be masked by a wrapper or iterator, or not called before the original response is thrown away (e.g. error cases). See e0c5ed19e25c77efe8944402353ff9b88abd1ae5. Thanks for pointing this out!

nayuki:
Amazing work Marcel, thank you!

Issue #1311, opened 2021-01-08
Accept pathlib paths across the board


Jwink3101:
This is an issue in preparation for a patch I will be trying to make.

I want to update Bottle to accept `pathlib.Path` paths across the board. My plan is to do this in a **backwards COMPATIBLE** way by adding something like `str(inputfile)` so that if you pass a native string or unicode, nothing happens but if you pass a pathlib object, it gets converted.

First question: Does this plan make sense? I use this in my tools so that I don't have to think about what is passed and it is explicitly converted

Second question: Where should I be looking? I think the goal should be as much coverage as possible but I would take better over perfect\*. Off the top of my head:

* lookup paths for SimpleTemplate
* Upload handling
* (maybe) config options

Anywhere else? 

Other thoughts?

\*It took the python standard library a few major versions to transition too.

oz123:
Can you post an example snippet of how this change would look?

defnull:
Sounds fine, but instead of `str()` I'd suggest to add a (private) `_strpath()` function that raises TypeError for value types that make no sense. Using `str()` would render sanity checks in other libraries useless, as for example `None` would be passed on as a string.

Jwink3101:
### re Private method

It makes sense to use a private `_strpath()` method, but there are some complexities in that. The main issue is that (for some reason) pathlib does not offer its own method to turn a path into a string. The method suggested method [in the docs](https://docs.python.org/3/library/pathlib.html) is to use `str()`. So you cannot look for a certain attribute to call to make it a string. So without directly importing it, you can't do `isinstance`

This is also further complicated by the fact that (as I found in trying to answer oz123's question below), there are instance checks for base-string. So instead I propose the following. Note that this is *NOT* done in the `if py3k` conditional to account for backports

```python
try:
    from pathlib import Path # 3.4+ or backport in python2
except ImportError:
    class Path(object):
        pass # dummy class. We never call it. Only used for instance checks
...

def _strpath(path):
    """Convert pathlib.Path objects"""
    if isinstance(path,Path):
        return str(path)
    return path 
```
Of course, the other option is to import pathlib and just do an instance check but that would break python2. I've seen in other PRs that you are hesitant to break python2 "just because". This is a reason to but it is also pretty minor. And then option 2.5 which is leave this open and don't make changes *until* we're already broken python2

### re example: 

Here is an example, but it actually shows the complexity. This is in the FileUpload class

```python
[...]

PR #1499, opened 2025-09-28
Support passing pathlib.Path in some functionalities

aisk:
Related to #1311, Support using `pathlib.Path` in:

- `load_config`
- `static_file`
- `lookup` fields in `BaseTemplate`

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -14,6 +14,7 @@ License: MIT (see LICENSE for details)
 """
 
 import sys
+from pathlib import Path
 
 __author__ = 'Marcel Hellkamp'
 __version__ = '0.14-dev'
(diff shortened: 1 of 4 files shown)


bottle.py
```diff
@@ -2773,6 +2781,11 @@ def static_file(filename, root,
         check or continue partial downloads) are also handled automatically.
     """
 
+    if isinstance(root, Path):
+        root = str(root)
```
defnull, line 2785: You can skip the type check and just call `str(var)` or `Path(var)` to get what you need. This may even be cheaper than the type check :)

bottle.py
```diff
@@ -3976,7 +3989,10 @@ def __init__(self,
         self.name = name
         self.source = source.read() if hasattr(source, 'read') else source
         self.filename = source.filename if hasattr(source, 'filename') else None
-        self.lookup = [os.path.abspath(x) for x in lookup] if lookup else []
+        if lookup:
```
defnull, line 3992: Same here, just call `str` on each value unconditionally. The check for an empty lookup variable could also be skipped, as a list comprehension over an empty iterator results in an empty list. This is not a hot path, so a little bit of overhead is fine, readability is more important.

aisk: Unfortunately, the default value of `lookup` is `None`, because of the common pitfall of Python's mutable types as default parameters. So we have to check if lookup is a falsy value here.

PR #1518, opened 2026-03-26
🤖 Keep multipart memory limit at parser default

ReinerBRO:
Summary:\n- keep Request.POST multipart parsing from reusing MEMFILE_MAX as the parser-wide memory limit\n- preserve MEMFILE_MAX as the per-file in-memory buffering threshold\n- add a regression test for two buffered uploads whose combined size exceeds 100KB but stays below the parser default\n\nValidation:\n- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest test/test_multipart.py test/test_environ.py -q\n- python -m py_compile bottle.py test/test_environ.py\n

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1367,8 +1367,7 @@ class BaseRequest:
         if not boundary:
             raise MultipartError("Invalid content type header, missing boundary")
         parser = _MultipartParser(self.body, boundary, self.content_length,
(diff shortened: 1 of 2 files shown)


defnull:
This would create a security issue and you did not follow agents.md instructions.

PR #1522, opened 2026-05-26
[AI] Fix multipart memory limit for buffered files

### LINKED ISSUE #1470: New MultipartParser fails on files above 102400 bytes
**Affected versions:** 0.13+ (tested in 0.13.2, issue is still present in code for 0.14)
Works in 0.12.25

The new MultipartParser released in [Version 0.13](https://bottlepy.org/docs/dev/changelog.html#release-0-13) raises a MultipartError("Memory limit reached.") when the total file size exceeds 102400 bytes. This is due to MEMFILE_MAX being passed in (possibly mistakenly?) as the memory limit on [line 1353](https://github.com/bottlepy/bottle/blob/master/bottle.py#L1353).

image

The [default value for this parameter](https://github.com/bottlepy/bottle/blob/master/bottle.py#L3113) would be 2 ** 20 i.e. ~1mb, which is still fairly low but more reasonable than 100kb. Ideally though this should be either dependent on actual memory available or configurable, perhaps via a kwarg to run().

puneetdixit200:
Fixes #1470.

This keeps `MEMFILE_MAX` as the per-file in-memory threshold for multipart uploads, but stops reusing that same 100 KiB value as the parser-wide total memory limit. The parser now uses its existing larger total-memory default while each individual file part still spools to disk once it exceeds `MEMFILE_MAX`.

I manually reviewed the diff and verified the regression path with:

- `python3 -m unittest test.test_environ.TestRequest.test_multipart_many_small_files_above_memfile_total`
- `python3 -m unittest test.test_environ test.test_multipart`
- `python3 -m unittest discover -t . -s test`
- `git diff --check HEAD~1..HEAD`

AI disclosure: assisted by OpenAI GPT-5. I reviewed the change and test output before submitting.

License: I explicitly license this contribution as public domain.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -1367,8 +1367,7 @@ class BaseRequest:
         if not boundary:
             raise MultipartError("Invalid content type header, missing boundary")
         parser = _MultipartParser(self.body, boundary, self.content_length,
-            mem_limit=self.MEMFILE_MAX, memfile_limit=self.MEMFILE_MAX,
-            charset=charset)
+            memfile_limit=self.MEMFILE_MAX, charset=charset)
 
         for part in parser.parse():
             if not part.filename and part.is_buffered():
--- a/test/test_environ.py
+++ b/test/test_environ.py
@@ -357,6 +357,24 @@ class TestRequest(unittest.TestCase):
         self.assertEqual(['value2', '万难'], request.forms.getall('field2'))
         self.assertTrue('field2' not in request.files)
 
+    def test_multipart_many_small_files_above_memfile_total(self):
+        """ Environ: POST can parse multiple files that exceed MEMFILE_MAX in total. """
+        payload = 'x' * (BaseRequest.MEMFILE_MAX // 2 + 1)
+        e = tools.multipart_environ(fields=[], files=[
+            ('file1', 'filename1.txt', payload),
+            ('file2', 'filename2.txt', payload),
+        ])
+        request = BaseRequest(e)
+        files = request.files
+
+        try:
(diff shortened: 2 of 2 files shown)


eeshsaxena (commented):
The goal (not rejecting two legitimate files whose sizes sum above `MEMFILE_MAX`) is right, but dropping the `mem_limit=self.MEMFILE_MAX` argument entirely has a side effect worth a second look.

`_MultipartParser.__init__` defaults `mem_limit=2 ** 20` and enforces it in the parse loop (`if part.size + mem_used > self.mem_limit: raise MultipartError("Memory limit reached.")`). So after this change the total in-memory bound for a multipart POST is no longer tied to `MEMFILE_MAX` at all -- it becomes the hardcoded 1 MB default.

Two consequences:

1. An app that *lowers* `MEMFILE_MAX` (e.g. a memory-constrained service setting it to a few KB) no longer has that reflected in the total buffered-multipart limit; it silently gets 1 MB instead.
2. `memfile_limit` (per-part spill-to-disk) and `mem_limit` (total in-memory) are genuinely different knobs, so coupling both to `MEMFILE_MAX` was the actual bug -- but the fix leaves `mem_limit` unconfigurable rather than giving it its own source.

Would it be cleaner to pass an explicit total limit that scales with the configured value, e.g. `mem_limit=max(self.MEMFILE_MAX, <sensible floor>)`, or expose a separate `MEMFILE_MAX_TOTAL`/`mem_limit` setting, so the total bound stays under the app`s control instead of reverting to the library default?

Issue #1205, opened 2020-01-29
Mounted application uses error handles from parent.


defnull:
In 0.13 a mounted application will use its own plugins, but other aspects (error handlers, hooks) are not separated correctly. Mounted applications should be mostly independant from the parent, as they were in 0.12

PR #1523, opened 2026-06-24
🤖 Fix error handler resolution in mounted applications (Issue #1205)

cgmark101:
### Description
Following the maintainer's feedback on the complexity and open design questions around modifying hook and error handler resolution for natively mounted applications, this PR document these limitations and details the recommended workaround (`app.wsgi`).

### Changes
- Updated the docstring of `Bottle.mount()` to note that hooks and custom error handlers of natively mounted sub-applications are bypassed.
- Added a new subsection to `docs/plugins/index.rst` ("Hooks, Error Handlers and Application Mounting") explaining the behavior and providing an example of mounting via the sub-application's WSGI interface (`app.wsgi`) for full isolation.

### AI Disclosure and Manual Review
This contribution was prepared with the assistance of the AI agent **Antigravity (using Gemini 3.5 Flash)**.
The documentation changes were manually proofread and verified for formatting correctness.

### License
I explicitly license all my contributions to this project as public domain.

### DIFF
--- a/bottle.py
+++ b/bottle.py
@@ -755,6 +755,17 @@ class Bottle:
             root of the child application via `/prefix` in addition to
             `/prefix/`, consider adding a route with a 307 redirect to the
             parent application.
+
+            .. note::
+                When mounting a :class:`Bottle` instance directly, hooks (e.g.
+                ``@hook('before_request')``) and custom error handlers registered
+                on the child application are not run. Instead, the parent
+                application's hooks and error handlers are used.
+
+                If you require complete isolation for the sub-application's hooks
+                and error handlers, mount it as a WSGI application by passing
+                its WSGI interface (e.g. ``child_app.wsgi`` or wrapping it as a
+                raw WSGI application) instead of the :class:`Bottle` instance.
         """
 
         if not prefix.startswith('/'):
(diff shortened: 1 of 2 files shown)


defnull:
This only solves half of the issue, as hooks are sill not considered. For a good and clean solution we need a more involved refactor, maybe turning the hook+error+cast behavior into a decorator so we can wrap the route callback with it (multiple times if needed). There is also the question of hooks should all fire, or only the inner-most hooks (as with plugins), for which I do not have a definite answer right now.

As a workaround you can always fall back to mounting app.wsgi which is the most isolated mounting method. Maybe this is not an implementation issue, but more a documentation issue?

cgmark101:
Hi! Thanks for the feedback. 

I completely agree that a proper runtime fix would require a much more involved refactoring of how hooks, errors, and cast operations interact. As suggested, I have reverted the code changes and instead updated the documentation to address this:

1. Added a note to the `Bottle.mount()` docstring explaining that hooks and custom error handlers on natively mounted sub-applications are bypassed, recommending the `.wsgi` interface if complete isolation is required.
2. Added a new subsection under `docs/plugins/index.rst` ("Hooks, Error Handlers and Application Mounting") detailing this behavior with a concrete code example.

I've force-pushed the new commits and updated the PR description. Let me know if this looks good to you!

PR #1527, opened 2026-07-09
🤖 Fix MultiDict.values()/items() to return proper dict views (#1113)

### LINKED ISSUE #1113: MultiDict.values - incorrect implementation in python 3
https://github.com/bottlepy/bottle/blob/9fb3b05846b33e508545a26e424785e151d02323/bottle.py#L2112

Assuming MultiDict supports `dict` protocol (which docstring suggests it does - ".., but behaves exactly like a normal dict.."), `.values` implementation is not correct. The returned object should be a view which, among other things, supports `len`. See https://docs.python.org/3/library/stdtypes.html?#dict-views

apoorvdarshan:
Fixes #1113.

### Root cause

`MultiDict.values()` and `MultiDict.items()` overrode the implementations inherited from `MutableMapping` with one-shot generator expressions. Those generators had no `len()` and were exhausted after one iteration, unlike normal dictionary views.

### Fix

Remove the legacy `values()` and `items()` overrides and use the standard implementations already provided by `MutableMapping`. `MultiDict.__getitem__()` returns the newest value for each key, so the inherited views preserve the existing value semantics while adding sizing, re-iteration, and normal view containment behavior.

The legacy `itervalues` and `iteritems` names remain available as aliases to the inherited implementations.

### Tests

Added `test_views_are_reiterable_and_sized` to verify that `keys()`, `values()`, and `items()`:

- support `len()`;
- can be iterated more than once; and
- reflect the newest value for each key.

The full suite passes on the minimum supported Python 3.9: **360 passed**.

This contribution is dedicated to the public domain.

Disclosure: prepared with AI assistance; manually reviewed and verified locally.


defnull:
`MutableMapping` already implements `items()` and `values()`. If `MutableMapping` has a compatible implementation in all supported python versions,w e can just remove the overrides and get the same effect. Those are probably old Python 2.x compatibility workarounds that never got removed.

PR #1534, opened 2026-08-31
fix(security,core): use constant-time digest comparison and fix header parsing edge cases

vjymisal0:
### Summary

1. **Security**: Replaced custom non-constant-time character comparison in `_lscmp` with `hmac.compare_digest(tob(a), tob(b))` to eliminate timing side-channel attacks on signed cookie decoding.
2. **HTTP Header Parsing**:
   - Fixed crash in `_parse_http_header` when processing bare attributes (valueless parameters like `foo; bar; baz`) by verifying `'=' in attr` before unpacking.
   - Fixed `parse_range_header` dropping suffix ranges (e.g. `bytes=10-, -10`) when preceded by standard optional whitespace.
3. **Core & Plugins**:
   - Fixed `FormsDict.__getattr__` raising `AttributeError` for unhandled dunder attribute lookups instead of `'super' object has no attribute '__getattr__'`.
   - Fixed `JSONPlugin` to properly respect per-route and app `json.enable` / `json.dump_func` configurations.
   - Fixed `BaseResponse.copy()` to preserve `response.body`.

### Testing
- `python -m unittest discover -t . -s test` -> **382 tests passed (OK)**.
- Added unit tests in `test/test_outputfilter.py`, `test/test_sendfile.py`, and `test/test_formsdict.py`.


defnull:
Way to much for a single commit, and the 'security' fix does not fix a security issue. I also see changes not explained in the PR that break expected behavior.

Issue #287, opened 2012-02-16
request for pluggable json encode/decode functions to process response


vaswadhati:
Hi there,

Luv bottle so far. Ran into an expected issue when dealing with json encoding that I have fixed for myself temporarily in my local copy.  

Couple of issues so far:
1/. It is perfectly alright to json encode a list (or a tuple) for that matter - in the pragmatic world of building Ajaxy web apps. The latest dev version (that I am using) checks if the response is a dict and if so applies json encoding on it. I patched mine to apply json encoding if it is a dict or a list. Not 100% sure about tuples so I left it at that.

2/. The response consists of data retrieved from a database and as expected json encoding tripped when it saw datetime and date types. There is a known way to override json package's dumps and loads methods. I took patch (that I had to write when using Py-rs) and just inserted again into my copy of bottle.

Overall, worked around the issue and I am in good shape to continue to use bottle. What nags me is that I had plans of extending my json patch to support direct encoding of database results from the database layer API such as straight dbapi, sqlalchemy - something along the lines of what I found at http://www.mail-archive.com/turbogears-commits@googlegroups.com/msg03926.html - which means I will have to keep patching bottle as new versions get released.

Ideal would be if I could pass json encode and decode functions when Bottle() instantiated - something along the lines of:

<code>
 myapi = Bottle(json_encode = myencoder, json_decode=mydecoder);
</code>

Or some thing along those lines or better. You get the idea.

Making an automatic decision to encode to json based on response type is a pretty significant and important design decision.  And I think should come with an option to allow the user to configure what gets encoded and how.

Thanks,

r-pluss:
I think that I'm missing something here:
I generally provide simplejson with a lambda expression for the default argument to handle things like datetime.datetime and datetime.date . This works well in other places. However, trying to use it with the above examples is not working for me, I still get errors akin to datetime.datetime(x,x,x,x,x) is not JSON serializable. 

```python
dt_handler = lambda obj:(obj.strftime('%Y-%m-%d %H:%M:%S') if isinstance(obj,datetime.datetime) else (obj.strftime('%Y-%m-%d') if isinstance(obj, datetime.date) else None))
app=Bottle(autojson=False)
app.install(JSONPlugin(json_dumps = lambda s: dumps(s, default= dt_handler)))
```

Any guidance?

eevleevs:
Is there any way to override also json_loads?

PR #1541, opened 2026-09-15
implement json.enable and json.dump_func configs

alisonatwork:
As discussed on #1539, the `json.enable` config - which in 0.13 deprecated and replaced the `autojson` application constructor arg - does nothing. This pull request updates `json.enable` so that it behaves as `autojson` used to: an on/off switch for dict-to-JSON-by-default behavior. When it is off, dictionaries will be treated as iterable, resulting in the same output as a list or tuple of the keys.

I've also implemented the `json.dump_func` config option, which can be used to replace the default (CPython or UltraJSON) `dumps`.

There is a bit business logic complexity with the custom serialization function, because `JSONPlugin` still allows passing a `json_dumps` function in its constructor. In practice the only way to usefully do that in 0.13 was to uninstall the default `JSONPlugin` and reinstall a new one or else reach around and overwrite the `json_dumps` property on the installed plugin then call `reset` to reapply it (similar to solution described in https://github.com/bottlepy/bottle/issues/287#issuecomment-4001810).

I think the best way to support backward-compatibility here is to ignore both the `json.enable` and `json.dump_func` configs if and only if the `JSONPlugin` had a custom `json_dumps` set on it (whether through constructor or property set). The downside of this is that if you install a custom `JSONPlugin` then you can't use config to enable or disable it per-route, but I am going to guess that this is not really a use case most people will ever have in real life.

I am not sure how deep to go with updating the docs to describe this behavior, so open the PR first to see. Maybe it doesn't need to be documented because the configs are basically enough and should be the standard way going forward?


defnull:
Looks good to me already. Tests, backwards compatibility, docs, all covered. Very nice! Since you force-pushed just 10 minutes ago I'll still wait a bit :)

alisonatwork:
Yeah, sorry about that force-push, I realized the commit was connected to my work email address vs my personal one. Not a big deal, since we use Bottle at work, but just for consistency I like to try keep the same email everywhere. If you are happy with the MR, feel free to merge 👍

